PR #26, opened 2016-04-05
fix MNOE-52

x4d3:
Impac Widget without data, create a link to /apps that cant' be resolved
for a dedfault enterprise project

### DIFF
--- a/src/app/index.config.coffee
+++ b/src/app/index.config.coffee
@@ -78,3 +78,14 @@ angular.module 'mnoEnterpriseAngular'
         create: "#{mnoHub}#{impacPrefix}/dashboards/:dashboard_id/widgets"
 
     ImpacRoutesProvider.configureRoutes(data)
+
+
+  #======================================================================================
+  # IMPAC-THEMING: Configuring colour theme, layout, labels, descriptions, and features
+  #======================================================================================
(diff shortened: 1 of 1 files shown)


x4d3:
@alexnoox could you please review ?

alexnoox:
@x4d3 We don't need to override any impac-angular library in this case. This impac-angular service is supposed to be used in a host app.

Have a look at how we used the ImpacRouteProvider service: https://github.com/maestrano/mno-enterprise-angular/blob/master/src/app/index.config.coffee#L66

You should then only pass to the service the key you want to override, have a look at the impac-angular documentation on how to do this with the ImpacAssetsProvider service: https://github.com/maestrano/impac-angular#impac-assets-provider-assetssvccoffee

It should look like something like this in index.config.coffee
```
.config (ImpacThemingProvider) ->
  options =
    # Options to override

  ImpacThemingProvider.configure(options)
```

alexnoox:
some docs has been updated recently about this (by @xaun): https://github.com/maestrano/impac-angular/pull/150/commits/11aac9bacf619686dabf6a2355ab5b1522baab27

x4d3:
@alexnoox thanks for your comment and explanation, I've corrected accordingly

alexnoox:
LGTM ;)

PR #28, opened 2016-06-29
[MNOE-91] Native possibility to set a background picture on the login page

claire00:


ouranos:
@alexnoox this variable is used by the `mno-enteprise` stylesheets.
Is it documented somewhere? It can be set to false to not  have a background.

alexnoox:
No it is not documented. The less variables should be refactored also.
It could be documented directly in the less files.

👍

PR #29, opened 2016-07-01
[MNO-98] Display App Pricing

winnietan:


src/app/views/marketplace/marketplace-app.html
```diff
@@ -106,6 +106,29 @@ <h2 class="media-heading">{{vm.app.name}}</h2>
           </div>
         </div>
 
+      <!-- Pricing -->
+      <div ng-show="vm.isPriceShown" class="row top-buffer-1">
```
alexnoox: A `ng-if` would be more adapted as this section will never be shown during the session if it is not enabled.

alexnoox: @winnietan I forgot to mention you ;)

winnietan: ok, I will try to change it. :)

alexnoox:
@winnietan Perfect thx

PR #31, opened 2016-07-05
[MNOE-99] Better favicon management

claire00:


alexnoox:
👍

ouranos:
@alexnoox Should we have some documentation in either mnoe or mnoe-angular explaining how to set up the favicon?

alexnoox:
@ouranos Yes we need more documentation . That should be done in the README. This can be another task.

PR #35, opened 2016-08-02
Add possibility to select marketplace currency

x4d3:
Changing the pricing.currency settings allow to select the default displayed currency in app prices.

### DIFF
--- a/src/app/views/marketplace/marketplace-app.html
+++ b/src/app/views/marketplace/marketplace-app.html
@@ -112,7 +112,7 @@
           <div class="well" ng-switch="vm.app.pricing_plans">
             <h2>Price</h2>
             <ul>
-              <li ng-repeat="plan in vm.pricing_plans()">
+              <li ng-repeat="plan in vm.pricing_plans">
                 <b>{{plan.name}}</b>: {{plan.description}}<br>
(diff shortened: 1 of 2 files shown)


x4d3:
@alexnoox Could you please review ?

src/app/views/marketplace/marketplace-app.html
```diff
@@ -111,14 +111,8 @@ <h2 class="media-heading">{{vm.app.name}}</h2>
...
-            <ul ng-if="vm.app.pricing_plans.AUD">
-              <li ng-repeat="plan in vm.app.pricing_plans.AUD">
-                <b>{{plan.name}}</b>: {{plan.description}}<br>
-                <span>{{plan.currency}} {{plan.price.value}} {{plan.factor}}</span>
-              </li>
-            </ul>
-            <ul ng-if="!vm.app.pricing_plans.AUD && vm.app.pricing_plans.default">
-              <li ng-repeat="plan in vm.app.pricing_plans.default">
+            <ul>
+              <li ng-repeat="plan in vm.pricing_plans()">
```
alexnoox: @x4d3 It is not a good behaviour to call a function in a `ng-repeat`, you should use an attribute.

PR #37, opened 2016-08-05
Update Impac version

hedudelgado:
@alexnoox  @ouranos 

I am not sure if it is right or not, I do the PR anyway. Please let me know if it is a problem, I am working in a private express project that has a bug and this has been fixed in this last release.

Thanks


alexnoox:
Hi @hedudelgado,

As there is a tild (~) before the version, you don't need to specify the minor version. The last version of impac! will be automatically retrieved by `rake mnoe:frontend:update`.
If you want to be sure to have the last versions in your build remove the tmp files before the build:
```
# Rebuild the admin-frontend
rm -rf tmp/build tmp/cache
bin/rake mnoe:frontend:update
```
Cheers,
Alex

PR #48, opened 2016-08-23
Fix addOn to add_on issue

hedudelgado:
@alexnoox @ouranos @x4d3 

This change has been done to solve the problem that the following PR will cause.

https://github.com/maestrano/mno-enterprise/pull/69


alexnoox:
LGTM

PR #50, opened 2016-09-01
Fix layout on mobile

alexnoox:


alexnoox:
@ouranos Review and merge :innocent:

PR #51, opened 2016-09-06
[MNOE-113] Display backend flash messages

ouranos:
Display flash messages from the backend in a toastr.
They're passed this way:
    `?flash={"msg":"An error message.","type":"error"}`


ouranos:
@BrunoChauvet @xaun Could you have a look?

@alexnoox @hedudelgado FYI

BrunoChauvet:
:+1:

PR #54, opened 2016-09-15
Config backward compatible with the mnoe backend

ouranos:
Avoid Unknown provider error when the frontend is newer than the
backend and using constants that are not defined in the backend config
file.


BrunoChauvet:
:cake:

x4d3:
YES YES YES

PR #55, opened 2016-09-15
[MNO-281] Add Intercom integration

ouranos:


BrunoChauvet:
:+1:

src/app/components/analytics/analytics.svc.coffee
```diff
@@ -0,0 +1,28 @@
+angular.module 'mnoEnterpriseAngular'
+  .service 'AnalyticsSvc', (MnoeCurrentUser, INTERCOM_ID, $window) ->
```
alexnoox: Shouldn't it be named `Intercom` or `MnoeIntercom` ?

alexnoox: Also the dependency order should be: `($window, MnoeCurrentUser, INTERCOM_ID)`

Order:
- Framework services
- Custom services
- Constants

It is from somewhere in this style guide: https://github.com/johnpapa/angular-styleguide/blob/master/a1/README.md

alexnoox:
LGTM

Sometime a good al' `$timeout` can be the solution.

PR #56, opened 2016-09-16
Update impac-angular routes config for kpis & alerting

xaun:
@cesar-tonnoir @ouranos


cesar-tonnoir:
LGTM - In the Jira card associated to the creation of a custom-express for impac dev, could you please add something like "remove impacprefix and normalize routing"

PR #58, opened 2016-09-22
[MNOE-165] Does not permit to install a second app if the app is not multi_instantiable

x4d3:
See https://github.com/maestrano/maestrano-hub/pull/153

### DIFF
--- a/src/app/components/mnoe-api/app-instances.svc.coffee
+++ b/src/app/components/mnoe-api/app-instances.svc.coffee
@@ -25,17 +25,13 @@ angular.module 'mnoEnterpriseAngular'
       # Workaround as the API is not standard (return a hash map not an array)
       # (Prefix operation by '/' to avoid data extraction)
       # TODO: Standard API
-      MnoeOrganizations.getCurrentId().then(
-        (selectedId) ->
-          MnoeApiSvc.one('organizations', selectedId).one('/app_instances').get().then(
-            (response) ->
-              # Save the response in the local storage
-              MnoLocalStorage.setObject(MnoeCurrentUser.user.id + "_" + LOCALSTORAGE.appInstancesKey, response.app_instances)
-              # Process the response
-              processAppInstances(response.app_instances)
-        )
+      MnoeApiSvc.one('organizations', MnoeOrganizations.selectedId).one('/app_instances').get().then(
+        (response) ->
+          # Save the response in the local storage
+          MnoLocalStorage.setObject(MnoeCurrentUser.user.id + "_" + LOCALSTORAGE.appInstancesKey, response.app_instances)
+          # Process the response
+          processAppInstances(response.app_instances)
       )
-
 
     # Process app instances to append them to the public variable
     processAppInstances = (appInstances) ->
--- a/src/app/components/mnoe-api/app-instances.svc.coffee
+++ b/src/app/components/mnoe-api/app-instances.svc.coffee
@@ -25,13 +25,17 @@ angular.module 'mnoEnterpriseAngular'
       # Workaround as the API is not standard (return a hash map not an array)
       # (Prefix operation by '/' to avoid data extraction)
       # TODO: Standard API
-      MnoeApiSvc.one('organizations', MnoeOrganizations.selectedId).one('/app_instances').get().then(
-        (response) ->
-          # Save the response in the local storage
-          MnoLocalStorage.setObject(MnoeCurrentUser.user.id + "_" + LOCALSTORAGE.appInstancesKey, response.app_instances)
-          # Process the response
-          processAppInstances(response.app_instances)
+      MnoeOrganizations.getCurrentId().then(
+        (selectedId) ->
+          MnoeApiSvc.one('organizations', selectedId).one('/app_instances').get().then(
+            (response) ->
+              # Save the response in the local storage
+              MnoLocalStorage.setObject(MnoeCurrentUser.user.id + "_" + LOCALSTORAGE.appInstancesKey, response.app_instances)
+              # Process the response
+              processAppInstances(response.app_instances)
+        )
       )
+
 
     # Process app instances to append them to the public variable
     processAppInstances = (appInstances) ->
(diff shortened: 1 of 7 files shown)


src/locales/zh.locale.json
```diff
@@ -163,7 +163,9 @@
   "mno_enterprise.templates.dashboard.marketplace.index.title": "市场",
   "mno_enterprise.templates.dashboard.marketplace.index.select_all": "全部",
   "mno_enterprise.templates.dashboard.marketplace.show.back": "返回",
+  "mno_enterprise.templates.dashboard.marketplace.show.conflicting_app": "You cannot install this App as it is incompatible with:",
```
ouranos: Wrong translation...

src/locales/id.locale.json
```diff
@@ -205,7 +205,9 @@
   "mno_enterprise.templates.dashboard.marketplace.index.title": "Pasar",
   "mno_enterprise.templates.dashboard.marketplace.index.select_all": "Semua",
   "mno_enterprise.templates.dashboard.marketplace.show.back": "Kembali",
+  "mno_enterprise.templates.dashboard.marketplace.show.conflicting_app": "You cannot install this App as it is incompatible with:",
```
ouranos: Wrong translation...

src/app/components/mnoe-api/app-instances.svc.coffee
```diff
@@ -25,14 +25,18 @@ angular.module 'mnoEnterpriseAngular'
       # Workaround as the API is not standard (return a hash map not an array)
       # (Prefix operation by '/' to avoid data extraction)
       # TODO: Standard API
-      MnoeApiSvc.one('organizations', MnoeOrganizations.selectedId).one('/app_instances').get().then(
-        (response) ->
-          # Save the response in the local storage
-          MnoLocalStorage.setObject(MnoeCurrentUser.user.id + "_" + LOCALSTORAGE.appInstancesKey, response.app_instances)
-          # Process the response
-          processAppInstances(response.app_instances)
+      MnoeOrganizations.getCurrentId().then(
```
alexnoox: Why do you call `MnoeOrganizations.getCurrentId()` here?

It is supposed to already be called in the layout controller.

src/app/views/marketplace/marketplace-app.controller.coffee
```diff
@@ -54,12 +71,34 @@ angular.module 'mnoEnterpriseAngular'
...
-          app = _.findWhere(response.apps, { slug: $stateParams.appId })
-          app ||= _.findWhere(response.apps, { id: parseInt($stateParams.appId) })
-          vm.initialize(app)
+          apps = response[0].apps
+          appInstances = response[1]
+          appId = parseInt($stateParams.appId)
+          appsPerNid = {}
+          appsPerNid[a.nid] = a for a in apps
+
+          app = _.findWhere(apps, { slug: $stateParams.appId })
```
alexnoox: Do you want to be able to use the slug in the url?

src/app/views/marketplace/marketplace-app.controller.coffee
```diff
@@ -54,12 +71,34 @@ angular.module 'mnoEnterpriseAngular'
...
+                not subCategory.multi_instantiable and  subCategory.name in names
+              )
+            )
+
+          vm.initialize(app, appInstance, conflictingAppInstance)
+
       )
 
+
+
```
alexnoox: Please remove those lines

src/locales/id.locale.json
alexnoox: If no translation is found it will automatically fallback to english.

src/locales/zh.locale.json
alexnoox: If no translation is found it will automatically fallback to english

src/app/views/marketplace/marketplace-app.html
```diff
@@ -36,10 +37,18 @@ <h2 class="media-heading">{{vm.app.name}}</h2>
             <div class="well text-center well-details">
               <div class="row">
                 <div class="col-xs-12 col-sm-4 col-sm-offset-4 col-md-8 col-md-offset-2">
-                  <div class="btn-fluid-wrapper">
-                    <a href="" ng-click="vm.provisionLink()" class="btn btn-warning btn-promo">
+                  <div class="btn-fluid-wrapper" ng-switch on="vm.appInstallationStatus()">
+                    <div ng-switch-when="CONFLICT" class="message">
+                      <i class="fa fa-exclamation-triangle"></i>
+                      {{ 'mno_enterprise.templates.dashboard.marketplace.show.conflicting_app' | translate }}
+                      <a ui-sref="home.marketplace.app({appId: vm.conflictingAppInstance.app_id})">{{vm.conflictingAppInstance.app_name}}</a>
```
alexnoox: Please use interpolated params (cf. https://angular-translate.github.io/docs/#/api/pascalprecht.translate.filter:translate)

alexnoox:
Can you also specify the JIRA in the title if there is one?
Cheers

src/app/views/marketplace/marketplace-app.html
x4d3: I would, but how do I use interpoloate params, if I want to put an html element, and not only text ?

alexnoox: In this case it is ok, forgot what is said ;)

alexnoox: A translator will have to use a demonstrative adjective: http://www.chineselanguageguide.com/grammar/demonstrative-adjective.asp

src/locales/zh.locale.json
x4d3: I've put the key so that we could translate later. And let in english so that it is not completly weird. I'll remove it.

src/app/views/marketplace/marketplace-app.controller.coffee
x4d3: it was already like that previously

alexnoox:
LGTM

PR #77, opened 2016-10-11
[impac-383] impac dhb printing display fixes

xaun:
Elements surrounding impac-angular must be print-hidden.

@alexnoox please review & merge


xaun:
@alexnoox why isn't this merged yet? :beer:

PR #81, opened 2016-10-14
[MNOE-217] Fix app page

ouranos:
App page was broken if the user had an app that was no longer available
on the marketplace.
Also refactored the code.


ouranos:
@alexnoox please review carefully ;)

@x4d3 FYI

alexnoox:
LGTM

PR #82, opened 2016-10-14
Add pricing details to dock settings

hedudelgado:
@ouranos @alexnoox


src/app/views/apps/modals/app-settings-modal.html
```diff
@@ -26,13 +26,23 @@
...
+        <hr>
+        <div ng-show="!!app.freeTrialEndAt">
+            Free Trial: {{ app.freeTrialEndAt }}
+        </div>
+        <div ng-show="!app.freeTrialEndAt && app.perUserLicence">
+          <ng-pluralize count="app.licencesCount" when="{'0': 'No users', '1': '1 user', 'other': '{{app.licencesCount | number}} users'}"
+             tooltip-placement="bottom" tooltip="This is the total number of people in your organization using {{app.name}}"></ng-pluralize>
+        </div>
+        <div ng-show="!app.freeTrialEndAt && !app.perUserLicence">Pricing details inside your app</div>
+      </div>
```
alexnoox: Please use i18n keys.

src/app/views/apps/modals/app-settings-modal.html
```diff
@@ -26,13 +26,23 @@
...
         <button ng-click="helper.addOnSettingLauch(app)" class="btn btn-warning">{{ 'mno_enterprise.templates.impac.dock.settings.access_add-on_settings' | translate }}</button>
       </div>
-
+      <!-- Billing Details -->
+      <div ng-show="helper.canDisplayBilling()">
+        <hr>
+        <div ng-show="!!app.freeTrialEndAt">
+            Free Trial: {{ app.freeTrialEndAt }}
+        </div>
+        <div ng-show="!app.freeTrialEndAt && app.perUserLicence">
```
alexnoox: Where are `app.freeTrialEndAt` and `app.perUserLicence` defined?
Are they supposed to come from the API? (In this case the variables would be snake cased)

hedudelgado:
Please @alexnoox , have a look at the changes.

Thank you very much for your comments.

src/app/views/apps/modals/app-settings-modal.coffee
```diff
@@ -28,6 +28,7 @@ angular.module 'mnoEnterpriseAngular'
     # ----------------------------------------------------------
     $scope.helper = {}
     $scope.helper.canDeleteApp = ->
+      console.log(app)
```
alexnoox: Please remove debug logs ;)

hedudelgado:
I cleaned it up @ouranos @alexnoox.

src/app/views/apps/modals/app-settings-modal.html
```diff
@@ -26,13 +26,23 @@
...
-
+      <!-- Billing Details -->
+      <div ng-show="helper.canDisplayBilling()">
+        <hr>
+        <div ng-show="!!app.free_trial_end_at">
+            {{ 'mno_enterprise.templates.impac.dock.settings.free_trial' | translate }}: {{ app.free_trial_end_at }}
+        </div>
+        <div ng-show="!app.free_trial_end_at && app.per_user_licence">
+          <ng-pluralize count="app.licences_count" when="{'0': 'No users', '1': '1 user', 'other': '{{app.licences_count | number}} users'}"
+             tooltip-placement="bottom" tooltip="This is the total number of people in your organization using {{app.name}}"></ng-pluralize>
```
alexnoox: Please i18n this too.
Ask @clemthenem, he worked on pluralisation and i18n ;)

src/app/views/apps/modals/app-settings-modal.html
```diff
@@ -26,13 +26,23 @@
...
       </div>
-
+      <!-- Billing Details -->
+      <div ng-show="helper.canDisplayBilling()">
+        <hr>
+        <div ng-show="!!app.free_trial_end_at">
+            {{ 'mno_enterprise.templates.impac.dock.settings.free_trial' | translate }}: {{ app.free_trial_end_at }}
+        </div>
+        <div ng-show="!app.free_trial_end_at && app.per_user_licence">
+          <ng-pluralize count="app.licences_count" when="{'0': '{{ &quot;mno_enterprise.templates.impac.dock.settings.no_users&quot; | translate }}', '1': '{{ &quot;mno_enterprise.templates.impac.dock.settings.1_user&quot; | translate }}', 'other': '{{app.licences_count | number}} {{ &quot;mno_enterprise.templates.impac.dock.settings.users&quot; | translate }}'}"
```
hedudelgado: Heys @alexnoox , this is the best solution I found to solve the problem of the i18n within the ng-pluralize. I have not checked with @clemthenem  as I thought this could be enough. Please confirm this is ok, otherwise I will look for another solution or ask clement straight away.

Thanks!

clemthenem: I would suggest you add the MessageFormat plugin, using the bower command:
`bower install angular-translate-interpolation-messageformat`

Then add it to your mno-enterprise-angular/src/app/index.config.coffee file:
`$translateProvider.addInterpolation('$translateMessageFormatInterpolation');`
This means that it will use default ng-translate functionalities and only load this plugin when required.

You can then add this translation in the en.json file:
`mno_enterprise.templates.impac.dock.settings.nb_of_users: '{num_users, plural, =0{No users} =1{1 user} other{# users}}';`
1. 'num_users' - The name of the variable you are using to test the number
2. 'plural' - The type of action you want to use ( it could also be 'gender' for example)
3. All the different cases (# replace the variable value in your translation).

Then you can simply use it like this in your template
`<p 
translate="PLURAL" 
translate-values="{ num_users: app.licences_count }" 
translate-interpolation="messageformat"></p>`

If you wanna play around with it I made this plunkr:
http://plnkr.co/edit/GyrLvacl7rk6GgsH0S6o?p=preview

Let me know if you have any problem.
Here is the doc: https://angular-translate.github.io/docs/#/guide/14_pluralization

hedudelgado:
@alexnoox I have added the solution that @clemthenem suggested. I must thank you by the way Clement, it was pretty clear and straight forward. 
Thanks a lot.

bower.json
```diff
@@ -23,6 +23,11 @@
     "impac-angular": "~1.5.0-rc1",
     "jquery": "~3.1.0",
     "less": "~2.5.3",
+    "angular-translate": "~2.9.1",
```
ouranos: These lines are duplicated

hedudelgado:
@alexnoox updated.

bower.json
```diff
@@ -46,6 +47,17 @@
         "less/font-awesome.less",
         "./fonts/*"
       ]
+    },
+    "moment": {
+      "main": [
+        "moment.js",
+        "locale/id.js"
```
ouranos: Missing a `,`

bower.json
```diff
@@ -46,6 +47,17 @@
         "less/font-awesome.less",
         "./fonts/*"
       ]
+    },
+    "moment": {
```
alexnoox: Why do you need to do this? We already import `angular-moment`.

hedudelgado: Hey @alexnoox 
In angular moment docs:
```
Usage
Include both moment.js and angular-moment.js in your application.
```
My input
This solution is out of the box and works by loading the different locales files we need in the project from the bower.json. The best part is that we can load all of them, or also only the ones we need. I specified id and ch to avoid having unnecessary extra files, but it can be extended anytime!
  "moment": {
       "main": [
         "moment.js",
         "locale/*.js"

Once it is loaded, by passing the locale we are using to moment in the run, or in the select language box for changing it.
```
moment.locale(TheLocaleWeAreUsing)
```
The language will be set globally and therefore propagated to every bit in the code where we are binding the result of moment. 
The default language is English, it does not need to be loaded. 
I tried with Chinese, french and Indonesian and seems to work perfectly fine, so it is very easy to extend.

bower.json
hedudelgado: 👍

x4d3:
@alexnoox @hedudelgado  Would it be possible to merge this ?

alexnoox:
@x4d3 No we can't merge this now.

This PR derived from its original goal (_[MNOE-206] Add pricing details to dock settings_) to i18n and l10n of dates on `mno-enterprise-angular`. Those features need to be addressed in 2 PRs, especially after the recent work of @ouranos on i18n.

The i18n and l10n solution with angular-moment looks good, it needs to be reworked a bit.
* You are using directly using `moment` (`moment.locale($scope.selectedLangKey)`) where you sould use the `amMoment` service (cf. https://github.com/urish/angular-moment#usage).
* In the bower file, I don't think we need to load `moment.js` as it is already loaded. We will need to load the moment locales though.

PR #87, opened 2016-10-25
[MNOE-223] Fix empty list of apps when clicking on current company

alexnoox:


alexnoox:
@ouranos To review and merge

PR #92, opened 2016-11-15
[MNOE-105] Change mno-enterprise-angular default theme colors

adamaziz15:


alexnoox:
For the logo, we should use a logo with a transparent background.

src/app/stylesheets/custom/button.less
```diff
@@ -66,3 +66,9 @@
     }
   }
 }
+
+.default-link {
+  .btn-warning {
+    background-color: @decorator-main-color;
```
alexnoox: Which button needs this customisation?

src/app/stylesheets/custom/custom-impac.less
```diff
@@ -0,0 +1,16 @@
+// This file includes style changes that overwrite some of the defaults in impac-angular
+
+.content.accounts-balance {
```
alexnoox: This will only modify the account balance widget.

Also I think we should have a look in the impac-angular project if there is no distinct background color vars for the widget header and body. If not, create those variables.

src/app/stylesheets/custom/accordion.less
```diff
@@ -35,7 +35,7 @@
 
   .panel-group .panel .panel-heading {
     padding: 0;
-    background-color: @dashboard-account-panel-bg-color;
+    background-color: @dashboard-account-tabs-bg-color;
```
alexnoox: `@dashboard-account-panel-bg-color` is the right name to me. If you need to change the color of this variable, do it in the variable.less.

src/app/stylesheets/theme.less
```diff
@@ -5,16 +5,16 @@
 // Main colors
 //-------------------------------------------------------------
 // Dashboard background color
-@bg-main-color:                  #aeb5bf;
+@bg-main-color:                  lighten(#F2EBBF, 10%);
```
alexnoox: Can you find/use the color hexa without the lighten function?

alexnoox:
Could you also change the colors in https://github.com/maestrano/mno-enterprise?
Cheers.

src/app/stylesheets/custom/button.less
alexnoox: Do we still need this button style?

alexnoox:
Everything looks good to me. Can you just answer on the custom button style?

adamaziz15:
The custom button style was for the "Add a new company" button in the company select box. I am looking now for a better way to do that. In the meantime I did come across something that might need to be taken out. Is there any use for the variables in lines 119 - 121 here https://github.com/adamaziz15/mno-enterprise-angular/blob/feature/105-colors/src/app/stylesheets/variables.less . There is also some css related to this but I don't see this class on any components https://github.com/adamaziz15/mno-enterprise-angular/blob/feature/105-colors/src/app/components/dashboard-company-selectbox/dashboard-company-selectbox.less lines 22 -32. @alexnoox

adamaziz15:
@alexnoox I removed the custom button style in button.less. I realized that the @brand- variables were responsible for what I was looking to change.

adamaziz15:
@alexnoox I changed some of the login vars to make the theme look better on the login page. I think this should be all set. I also took a look at mnoe and it seemed that changing the logo on the login page was all that needed to be done. Please review, I also made the PR in mnoe.

alexnoox:
@adamaziz15 Very good catch for the variables.less lines 119 - 121 and the associated class, it is not used anymore. You can remove it.

alexnoox:
@adamaziz15 Thanks for this, I will check and merge it 👍

adamaziz15:
@alexnoox I removed those unused vars and the corresponding class. 👍

PR #105, opened 2016-12-14
average rating not updating

hedudelgado:
@alexnoox @ouranos

### DIFF
--- a/src/app/components/mnoe-api/marketplace.svc.coffee
+++ b/src/app/components/mnoe-api/marketplace.svc.coffee
@@ -20,7 +20,7 @@ angular.module 'mnoEnterpriseAngular'
 
     @addAppReview = (appId, data) ->
       payload = {app_review: data}
-      MnoeApiSvc.one('marketplace', parseInt(appId)).doPOST(payload, '/app_review').then(
+      MnoeApiSvc.one('marketplace', parseInt(appId)).post('/app_review', payload).then(
         (response) ->
           app_review = response.plain()
           app_review
--- a/src/app/components/mnoe-api/marketplace.svc.coffee
+++ b/src/app/components/mnoe-api/marketplace.svc.coffee
@@ -20,7 +20,7 @@ angular.module 'mnoEnterpriseAngular'
 
     @addAppReview = (appId, data) ->
       payload = {app_review: data}
-      MnoeApiSvc.one('marketplace', parseInt(appId)).post('app_review', payload).then(
+      MnoeApiSvc.one('marketplace', parseInt(appId)).doPOST(payload, '/app_review').then(
         (response) ->
           app_review = response.plain()
           app_review
--- a/src/app/views/marketplace/marketplace-app.controller.coffee
+++ b/src/app/views/marketplace/marketplace-app.controller.coffee
@@ -42,11 +42,11 @@ angular.module 'mnoEnterpriseAngular'
         vm.isLoading = false
 
       # Update the comments and avergae review
-      vm.updateReview = (newReview) ->
+      vm.updateReview = (response) ->
         # Select and push to comments the last comment added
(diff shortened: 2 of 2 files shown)


src/app/components/mnoe-api/marketplace.svc.coffee
```diff
@@ -20,7 +20,7 @@ angular.module 'mnoEnterpriseAngular'
 
     @addAppReview = (appId, data) ->
       payload = {app_review: data}
-      MnoeApiSvc.one('marketplace', parseInt(appId)).post('app_review', payload).then(
+      MnoeApiSvc.one('marketplace', parseInt(appId)).doPOST(payload, '/app_review').then(
```
ouranos: Why is that? Is the `post` only for classic CRUD action?

src/app/views/marketplace/marketplace-app.controller.coffee
```diff
@@ -41,6 +41,14 @@ angular.module 'mnoEnterpriseAngular'
 
         vm.isLoading = false
 
+      # Update the comments and avergae review
+      vm.updateReview = (newReview) ->
+        # Select and push to comments the last comment added
+        [first, ..., last] = newReview.app.reviews
```
ouranos: Please review variable name: `newReview.app.reviews` is confusing

I guess it's `response.app.reviews`

ouranos: Also replace
`[first, ..., last] = newReview.app.reviews`
by
`[..., newReview] = response.app.reviews` 

And use `newReview` instead of `last`

src/app/components/mnoe-api/marketplace.svc.coffee
hedudelgado: when doing post we are not able to get the whole app object in the response, but a simple review, regardless all the magic we do on the back end, it is due the mnoApiSvc too, cause the, and it is also the way is done in other parts of the project.

api.svc.coffee
```
          if (_.startsWith(what, '/'))
            extractedData = data
          # On getList extract and restangularize the objects list
          else if (operation == 'getList')
            extractedData = data[what]
          # Extract and restangularize the object
          else if (operation == 'get' || operation == 'post' || operation == 'put')
            what = inflector.singularize(what)
            extractedData = data[what]
          else
            extractedData = data
          return extractedData
      )
```

Due the fact we are getting the whole object instead of the single review with the regular post, the object in angular will be undefined.
            extractedData = data[what]

alexnoox: Have you tried something like this:
```
MnoeApiSvc.one('marketplace', parseInt(appId)).post('/app_review', payload).then(...)
```

I added a `/`, in `app_review`

hedudelgado: @alexnoox ,thanks a lot. You can check that there is a new commit with that change added.

src/app/views/marketplace/marketplace-app.controller.coffee
```diff
@@ -41,6 +41,14 @@ angular.module 'mnoEnterpriseAngular'
 
         vm.isLoading = false
 
+      # Update the comments and avergae review
+      vm.updateReview = (response) ->
+        # Select and push to comments the last comment added
+        [first, ..., newReview] = response.app.reviews
```
ouranos: As mentioned in the previous comment, you don't need `first`, so you can decompose the array this way:
`[..., newReview] = response.app.reviews`

hedudelgado:
@ouranos I changed and test it. It works perfectly fine. 
Thanks a lot

ouranos:
@alexnoox I believe this one can be closed as we did in #106 
We no longer the full list of reviews but just the new one and the average rating

alexnoox:
Yes, this can be closed.

PR #106, opened 2016-12-16
Add pagination

hedudelgado:
@alexnoox


alexnoox:
Could you add a # of elements per page?

Like this, on the right of the pagination: 
screen shot 2016-12-19 at 12 02 04 pm

hedudelgado:
@alexnoox

ouranos:
@hedudelgado reworked the pagination to make it work

@alexnoox a bit of polish to do on the frontend

This requires my changes on mno-enteprise that I'm yet to push

PR #112, opened 2016-12-23
Fix angular token error

adamaziz15:
@alexnoox I noticed a token error in the console anytime there is an in-app alert. This was due to the ng-bind-html directive having braces => {} around it.


alexnoox:
I've been testing it in a plunker (https://plnkr.co/edit/R7UTXmZnG31Z2CbrqknB?p=preview) and you're right.

Thank you for pointing this 🎅

PR #115, opened 2017-01-04
Fix bower angular dependency resolutions

alexnoox:


alexnoox:
@ouranos Please review

@cesar-tonnoir FYI

PR #122, opened 2017-01-12
App rating

ouranos:
@alexnoox could you review the last 3 commits?
- Fix sanitization (I reverted the replacement of a copy by  an assignment, not sure of side effects)
- Fix display when no rating
- Feature flag

Full sanity review would be good too ;)


src/locales/en.locale.json
```diff
@@ -441,6 +442,18 @@
...
+  "mno_enterprise.templates.dashboard.marketplace.show.on": "on",
+  "mno_enterprise.templates.dashboard.marketplace.show.just_now": "Just Now",
+  "mno_enterprise.templates.dashboard.marketplace.show.number_of_reviews": "Reviews per page:",
+  "mno_enterprise.templates.dashboard.marketplace.show.company_feedback": "Company feedback",
+  "mno_enterprise.templates.dashboard.marketplace.show.review": "Review",
+  "mno_enterprise.templates.dashboard.marketplace.show.tell_others": "Tell others what you think about this application (optional)",
+  "mno_enterprise.templates.dashboard.marketplace.show.cancel": "Cancel",
+  "mno_enterprise.templates.dashboard.marketplace.show.submit": "Submit",
+  "mno_enterprise.templates.dashboard.marketplace.show.success_toastr": "Your review has been taken into account.",
+  "mno_enterprise.templates.dashboard.marketplace.show.error_toastr": "An error occurred while saving your review.",
```
alexnoox: We need to make sure that all the locales are in `mno-enterprise`.

ouranos: I double checked. They're all there: https://github.com/maestrano/mno-enterprise/pull/186/files#diff-cecaaa6d713134a5acc8e508b43e87d1R23

PR #133, opened 2017-01-24
Fix team view: add an app button & message color

alexnoox:


alexnoox:
@ouranos Please review

@fgourichon FYI

PR #138, opened 2017-01-26
WIP: [MNOE-282] Comparison of apps in the Marketplace

MichaelSedov:


src/app/views/marketplace/marketplace.controller.coffee
```diff
@@ -1,41 +1,70 @@
...
+        count = 0
+        angular.forEach vm.apps, (value) ->
+          if value.is_compare
+            count++
+          return
+        count
+
+      # Comparison function
+      vm.Comparison = (event) ->
+        event.preventDefault()
```
alexnoox: Do you really need to prevent the default event here?

src/app/views/marketplace/marketplace.controller.coffee
```diff
@@ -1,41 +1,70 @@
...
+        checkedApp = vm.calculateChecked()
+        if (checkedApp > 4 || checkedApp < 2)
+          toastr.error 'Need to select 2 to 4 items', 'Comparison'
+        else
+          $state.go('home.marketplace.compare')
+
+      # Uncheck all checkboxes
+      vm.unCheckedApps = () ->
```
alexnoox: Could be named `vm.uncheckAllApps`.

src/app/views/marketplace/marketplace.controller.coffee
MichaelSedov: This stops the transfer comparison page if validation fails

PR #153, opened 2017-02-01
Added explanation when no reviews and added some style

clemthenem:


src/app/views/marketplace/marketplace-app.html
```diff
@@ -135,15 +135,18 @@ <h2 class="media-heading">{{vm.app.name}}</h2>
               {{ 'mno_enterprise.templates.dashboard.marketplace.show.write_review' | translate }}
             </button>
             <span class="pull-right">
-              {{ 'mno_enterprise.templates.dashboard.marketplace.show.filter' | translate }}
-              <select ng-model="vm.sortReviewsBy" ng-change="vm.orderFeedbacks()">
+              <label>{{ 'mno_enterprise.templates.dashboard.marketplace.show.filter' | translate }}</label>
+              <select class="filter-reviews form-control" ng-model="vm.sortReviewsBy" ng-change="vm.orderFeedbacks()">
```
alexnoox: `select` should have an `id` attribute and `label` should have a `for` attribute.

PR #163, opened 2017-02-20
[MNOE-339] - Fix search issue on questions

clemthenem:
@alexnoox can you review this please? :)


alexnoox:
I'm not sure about this one.

It seems that you try to search in the question `description` and also in the question `answers`. I think if you type `id` this will return all the answers as they all contain and `id` key.

You may have to create your own filter, checking the `description` and then filtering the answers `description`.

Have a look at this syntax: http://stackoverflow.com/questions/19405346/how-can-i-write-an-angularjs-filter-to-search-for-string-fragments-sequences-of

clemthenem:
@alexnoox I changed this with a filter function!
Let me know if that does not suit you

PR #170, opened 2017-03-03
[MNOE-357] - Error message displayed incorrect

clemthenem:
@ouranos


alexnoox:
@ouranos Shouldn't this be merged also in master?

PR #173, opened 2017-03-09
[MNOE-379] User & Organization management feature flag

ouranos:


ouranos:
See https://github.com/maestrano/mno-enterprise/pull/254

PR #177, opened 2017-03-22
Fix app install button component behaviour

alexnoox:
@ouranos Please review and merge


ouranos:
I think we should rebase and merge on `1.0`

PR #190, opened 2017-05-09
Add URL_CONFIG

fgourichon:
@ouranos for review


src/app/index.route.coffee
```diff
@@ -39,16 +39,22 @@ angular.module 'mnoEnterpriseAngular'
...
+
             if I18N_CONFIG.enabled
-              $window.location.href = "/#{$translate.use()}#{URI.login}"
+              if URL_CONFIG.after_sign_out_url
+                $window.location.href = "#{logout_url}"
+              else
+                $window.location.href = "/#{$translate.use()}#{URI.login}"
             else
-              $window.location.href = "#{URI.login}"
+              $window.location.href = "#{logout_url}"
```
ouranos: Any reason why we had it before?

src/app/index.route.coffee
```diff
@@ -39,16 +39,22 @@ angular.module 'mnoEnterpriseAngular'
...
             AnalyticsSvc.logOut()
+
+            logout_url = URL_CONFIG.after_sign_out_url || "#{URI.login}"
+
             if I18N_CONFIG.enabled
-              $window.location.href = "/#{$translate.use()}#{URI.login}"
+              if URL_CONFIG.after_sign_out_url
+                $window.location.href = "#{logout_url}"
+              else
+                $window.location.href = "/#{$translate.use()}#{URI.login}"
```
ouranos: It's the same as the previous code.

alexnoox: `URI.login` has a `/` as first character

alexnoox: all good ;)

src/app/index.route.coffee
alexnoox: No, but I suppose it could be removed.

fgourichon:
Removed interpolations

PR #200, opened 2017-05-26
[MNOE-484] Fix locale interpolation problem

alexnoox:
Better compatibility with impac-angular 1.5

@ouranos Please review


alexnoox:
https://github.com/maestrano/mno-enterprise/pull/303 contains locales in MNOE

PR #202, opened 2017-06-05
[MNOE-464] - Remove connect button for non admin users

fgourichon:
@ouranos and @alexnoox for review


src/app/components/mnoe-api/organizations.svc.coffee
```diff
@@ -230,4 +230,8 @@ angular.module 'mnoEnterpriseAngular'
       organizationSettings: (obj = null) -> _self.can.create.organizationSettings(obj) # call similar permission
     }
 
+    _self.can.connect = {
+      appInstance: (obj = null) -> _self.role.atLeastAdmin()
```
alexnoox: Should be `appInstance: -> _self.role.atLeastAdmin()`

fgourichon: Indeed. Fixed

PR #209, opened 2017-06-12
[MNOE-473] Split review/question feature flag

adamaziz15:
@alexnoox @ouranos I've split the review feature flag so there is the ability to have reviews without questions/answers. Currently, you cannot have questions/answers without reviews. Is that something we should have the ability to do as well?


alexnoox:
The code LGTM.

For your question, I think that it is something that may be needed.

adamaziz15:
@alexnoox Updated to allow for questions without reviews as well 👍

PR #212, opened 2017-06-13
[MNOE-415] Remove editing app name

hedudelgado:


hedudelgado:
done @ouranos , sorry

PR #218, opened 2017-06-15
[MNOE-486] Fix pagination when more than 5 comments

hedudelgado:
@alexnoox

### DIFF
--- a/src/app/components/mno-pagination/mno-pagination.html
+++ b/src/app/components/mno-pagination/mno-pagination.html
@@ -9,7 +9,7 @@
 </div>
 
 <ul uib-pagination ng-model="page" 
-  ng-show="totalItems > nbItems" 
+  ng-show="totalItems > 0" 
   total-items="totalItems"
   items-per-page="nbItems"
   ng-change="onPageChange()"
--- a/src/app/components/mno-pagination/mno-pagination.html
+++ b/src/app/components/mno-pagination/mno-pagination.html
@@ -9,7 +9,7 @@
 </div>
 
 <ul uib-pagination ng-model="page" 
-  ng-show="totalItems > 0" 
+  ng-show="totalItems > nbItems" 
   total-items="totalItems"
   items-per-page="nbItems"
   ng-change="onPageChange()"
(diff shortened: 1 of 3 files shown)


src/app/components/mno-pagination/mno-pagination.html
```diff
@@ -9,7 +9,7 @@
 </div>
 
 <ul uib-pagination ng-model="page" 
-  ng-show="totalItems > nbItems"
```
alexnoox: This line was preventing the pagination to be displayed if not needed, `totalItems > 0` will always display it?

src/app/views/marketplace/marketplace-app.controller.coffee
```diff
@@ -169,6 +169,7 @@ angular.module 'mnoEnterpriseAngular'
           (response) ->
             vm.reviews.list.splice(key, 1)
             updateAverageRating(response.average_rating)
+            vm.reviews.pageChangedCb(vm.app.id, vm.reviews.nbItems, vm.reviews.page = 1)
```
alexnoox: This will refresh the list calling the backend. To have a faster UI, we could use used `splice` or `unshift` methods.

ouranos:
Hey @hedudelgado, how does that fit with #225. Should we have them both in the same PR? Thanks

hedudelgado:
Hey @ouranos , they are closely related, but as one is for the reviews, and the other for questions, I open two different PRs.

src/app/views/marketplace/marketplace-app.controller.coffee
hedudelgado: @alexnoox , using unshift or splice cause wrong workflow, as it does not team up with pagination at all:

-Being in any page that is not first, and adding a question or review, it will show as the first one on that page instead of in page 1.
-When deleting the last comment/question in pages > 1, it will show a message "no reviews/questions", when there are still in other pages. 

This is the only reasonable solution I found. Since a user, normally would post/delete once, it won't over hit the backend.

Thoughts?

hedudelgado:
@alexnoox as mentioned in Jira, conflicts fixed on this pr, you can also have a look at the conversation thread we had.

PR #224, opened 2017-06-16
[MNOE-496] Fix wrong behaviour on install app button

hedudelgado:


alexnoox:
If the app is initialised using: `vm.app = {}`, the `ng-if` condition will be true all the time, no?

In the marketplace app view, the `vm.app` variable should only be initialised in the init phase (when the app is fetched) or set by default to `null`.

hedudelgado:
Hey @alexnoox ,
Taking into account that object = {} is true, I got curious in understanding why this actually works with "ng-if" and it does not with "ng-show".

Works because:
ng-if executes at priority level 600.
ng-show executes at priority level 0.
ng-controller executes at priority level 500.

Greater number compile first, so: 
1) ng-if evaluates undefined object in digest 1
2) controller compiles, initialize and set vm.app equal to actual app in digest 2
3) ng-if detect that vm.app is now defined and then compiles the button with a real app in digest 3

In the other hand, since we are using **angular.copy(app, vm.app)**, we can not pre-initialize the object with null, we need to initialize vm.app ={}, otherwise it will return null.

alexnoox:
Thanks for your response

PR #228, opened 2017-06-21
Feature/transac add reconcile view

xaun:
@alexnoox please review (don't merge, just wanting to open up these branches to start getting them review / planning for merge / release / tag etc)

https://github.com/maestrano/transac/pull/1

FYI @BrunoChauvet


src/app/index.config.coffee
```diff
@@ -8,7 +8,7 @@ angular.module 'mnoEnterpriseAngular'
     # Set options third-party lib
     toastrConfig.timeOut = 5000
     toastrConfig.positionClass = 'toast-top-right'
-    toastrConfig.preventDuplicates = true
+    toastrConfig.preventDuplicates = false
```
xaun: Noox is there a reason this is set to true in the framework? it prevents alerting non-unique messages.

alexnoox: The reason is to prevent duplicates :trollface: 
If you need it for Transac!, we can set it to `false`.

xaun: Obviously nooooox :P But yeah it's not super crucial, but I was using it in Transac!

PR #234, opened 2017-06-27
Default variables: change less files priority & fix layout

alexnoox:
@ouranos Please review


src/app/index.less
```diff
@@ -14,8 +14,11 @@
 @font-path: '/admin/fonts';
 
 html, body {
+  height: 100%;
```
cesar-tonnoir: ok as it will be overridden by `101% important!` :p

PR #236, opened 2017-06-27
[MNOE-497] [WIP] Move rails pages to SPA

adamaziz15:
@alexnoox @ouranos


src/app/views/auth/auth.less
```diff
@@ -149,3 +149,272 @@
     height: 34px;
   }
 }
+
+input.fluroblue[type=radio] {
```
alexnoox: `fluroblue` is color specific to maestrano, please use an agnostic selector name and remove those occurences.

src/app/index.config.coffee
```diff
@@ -25,13 +25,15 @@ angular.module 'mnoEnterpriseAngular'
...
+    $httpProvider.interceptors.push ($q, $window, $injector, $log) ->
       {
         responseError: (rejection) ->
           if rejection.status == 401
             # Redirect to login page
-            console.log "User is not connected!"
-            $window.location.href = '/'
+            toastr = $injector.get('toastr')
+
+            toastr.error('User is not connected!')
```
alexnoox: We should i18n this toastr

src/app/index.config.coffee
```diff
@@ -69,3 +71,22 @@ angular.module 'mnoEnterpriseAngular'
...
+
+
+
+
+
+
+
+
+
+
```
alexnoox: Remove empty lines

src/app/index.config.coffee
```diff
@@ -69,3 +71,22 @@ angular.module 'mnoEnterpriseAngular'
     # TODO: Activate in "developer mode" only (spams the console and makes the application lag)
     #$translateProvider.useMissingTranslationHandlerLog()
   )
+  # Configure auth routes
+  .config((AuthProvider, AuthInterceptProvider) ->
+    AuthProvider.loginPath('/mnoe/auth/users/sign_in')
```
alexnoox: Please use a constant to define those path strings

PR #241, opened 2017-07-04
Api V2 - Adapt code for Reviews

x4d3:


ouranos:
@alexnoox please review. We'll need to merge maestrano/mno-enterprise#349 first though.

PR #250, opened 2017-07-11
Add gulp task to ensure locales in development

MarcoCode:
- Remove deprecated `util.isArray()`

@ouranos @cesar-tonnoir @xaun 
Tiny gulp task that runs before `gulp serve` to ensure locales are available in development

### DIFF
--- a/gulp/locales.js
+++ b/gulp/locales.js
@@ -1,0 +1,12 @@
+'use strict';
+
+var gulp   = require('gulp');
+var $ = require('gulp-load-plugins')();
+
+gulp.task('locales', function () {
+  gulp.src('bower_components/impac-angular/dist/locales/*')
+  .pipe($.rename(function (filename) {
+    filename.basename = filename.basename.slice(0, 2);
+  }))
+  .pipe(gulp.dest('src/locales/impac'))
+});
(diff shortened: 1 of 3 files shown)


gulp/locales.js
```diff
@@ -0,0 +1,12 @@
+'use strict';
+
+var gulp   = require('gulp');
+var rename = require('gulp-rename');
```
xaun: Is rename accessible through this?
```
var $ = require('gulp-load-plugins')();

$.rename()
```
I can see in the mnoe-angular `gulp/inject.js` the script is accessing libs via `$`, but it has also defined the rename lib to a variable, so not sure whether this is done for a reason. If not, could be refactored there too.

gulp/locales.js
```diff
@@ -0,0 +1,12 @@
...
+
+gulp.task('locales', function () {
+  gulp.src('bower_components/impac-angular/dist/locales/*.*')
```
xaun: Could this glob just be `/*` ?

MarcoCode: 💯

alexnoox:
Very cool guys ;)

PR #253, opened 2017-07-13
[MNOE-544] Onboarding: fixes

alexnoox:


alexnoox:
Locales: https://github.com/maestrano/mno-enterprise/pull/370

PR #254, opened 2017-07-13
[DEVPL-158] Pricing sync

adamaziz15:


ouranos:
We shold add some guarding to ensure backward compatibility with older MnoHub or Mnoe
@alexnoox is `ng-bind-html` the best choice here?

alexnoox:
@ouranos `ng-bind-html` is already used for `vm.app.description`.

alexnoox:
If `vm.app.pricing_text` is `undefined`, nothing will be displayed so no need of guarding.

PR #233, opened 2017-06-27
[MNOE-501] UI improvement on-boarding

rheasunshine:
@ouranos Please review

PR #255, opened 2017-07-20
Revert "[MNOE-501] UI improvement on-boarding"

alexnoox:
Reverts maestrano/mno-enterprise-angular#233


alexnoox:
I shouldn't have merged, there is a mismatch in the locale keys name.

PR #263, opened 2017-07-26
[MNOE-474] Fix reviews for API v2

ouranos:


ouranos:
@alexnoox rebased and removed the `parseInt`

PR #268, opened 2017-07-31
[MNOE-558] Update handling of admin_role for reviews/questions frontend

adamaziz15:
@alexnoox 

Edit: 
FYI the JIRA description:
> Issue 2: When a comment is edited by admin panel, "edited by user" is appearing instead of "platform moderator"
> Issue 3 : When answer is posted by "admin" it is appearing as "Posted by user"

### DIFF
--- a/src/app/views/marketplace/marketplace-app.html
+++ b/src/app/views/marketplace/marketplace-app.html
@@ -209,10 +209,10 @@
                       <span ng-if="comment.created_at"> {{comment.created_at | date: "dd/MM/yyyy" }}</span>
                       <span ng-if="!comment.created_at"> {{ 'mno_enterprise.templates.dashboard.marketplace.show.just_now' | translate }}</span>
                       <span ng-if="comment.edited"> -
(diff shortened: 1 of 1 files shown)


src/app/views/marketplace/marketplace-app.html
```diff
@@ -264,7 +264,8 @@ <h2 class="media-heading">{{vm.app.name}}</h2>
                       </div>
                       <div class="col-md-10">
                         <p class="question-text">{{ question.description }}</p>
-                        <p class="question-attribute">Posted by: {{ question.user_name }} in {{ question.organization_name }} on {{ question.created_at | date: "dd/MM/yyyy" }}</p>
+                        <p ng-hide="question.user_admin_role == 'admin'" class="question-attribute">Posted by: {{ question.user_name }} in {{ question.organization_name }} on {{ question.created_at | date: "dd/MM/yyyy" }}</p>
```
ouranos: @alexnoox For consistency should we use `ng-if`/`ng-show` for both comment and questions? Which one is better?

alexnoox: In this case it seems it is only displayed if we are an admin, so use `ng-if`.

ouranos:
@alexnoox after discussion I've switched to `ng-if`, please review and merge

PR #270, opened 2017-07-31
[WIP] [MNOE-551] Tasks

xaun:
hey @alexnoox, I am working on the Tasks and I am wanting to involve your reviews as I go along as the timeline is very tight. 

I am currently building out the UI components stubbed, and will connect to the API in the following couple of days. 

I am trying to leverage mno-ui-elements by creating components as I go (accompanying PR: https://github.com/maestrano/mno-ui-elements/pull/4).  

But, for example, the `create-task-modal` is already using ui-bootstrap for so many things there isn't really anything I can move to mno-ui-elements... So I am wondering whether I should've made a separate lib for this... As there will be duplication in mnoe-admin-panel.

Can you create a `feature/tasks` branch on upstream and switch this PR to that?

Thanks!


src/locales/id.locale.json
```diff
@@ -202,6 +202,7 @@
   "mno_enterprise.templates.dashboard.menu.marketplace": "Pasar",
   "mno_enterprise.templates.dashboard.menu.logout": "Keluar",
   "mno_enterprise.templates.dashboard.menu.new_company": "Perusahaan Baru",
+  "mno_enterprise.templates.dashboard.menu.messages": "Pesan",
```
alexnoox: You don't need to add a translation in those files. The default fallback language is `en.json`

xaun:
Hey Noox, thanks for the review.

> Why don't you simply move the component create-task-modal into mno-ui-element lib, if you are going to use it in mno-enterprise-angular and mnoe-admin-panel?

Oh ok, I wasn't sure how specific the mno-ui-elements could be. 

> You can create a component wrapping smart-table

Smart table looks ok, can you customise the sorting arrows? can't find something on this and the default ones are ugly. Also, in the demos provided, the column sorting feels slow compared to the native angular sort I have done. Have you noticed this?

I have made the mnoSortableTable component pretty generic, do you like this? Or should I just create a mnoTasks component that is more specific?

> There is a lot of LESS styles that I can't find used in templates.

Not sure what less styles you're talking about here?

Cheers

alexnoox:
About the smart table:
- Yes you can configure the sorting arrows (cf. https://github.com/maestrano/mnoe-admin-panel/blob/d281dc50fa130e1dc457ea2dbe90c5808a5536de/src/app/stylesheets/smart-table.less)
I think you will be able to have the same one you designed, I like how they look btw.
- I think `mnoSortableTable` is great because it is simple. Of course, it might be faster now. But don't forget that you will need to interrogate the server using callbacks (cf. http://lorenzofox3.github.io/smart-table-website/#section-pipe), add loaders, pagination, search, etc. and you will end up re-developing smart-table. Smart-Table is a mature project that we already manipulated and work well. (cf. https://github.com/maestrano/mnoe-admin-panel/blob/2.0/src/app/components/mnoe-staffs-list/)

PR #277, opened 2017-08-08
[MNOE-455] Impersonation - Add user consent

x4d3:


src/app/components/impersonification-notification/impersonification-notification.html
```diff
@@ -0,0 +1,22 @@
+<div class="container" ng-show="requests.length > 0">
```
alexnoox: There should only be one `.container` in a bootstrap layout. There is already one in `layout.html`, so you can remove this one.

src/app/components/impersonification-notification/impersonification-notification.html
```diff
@@ -0,0 +1,22 @@
...
+  <div class="row">
+    <div class="col-lg-10 col-lg-offset-1">
+      <div class="well">
+        <mno-loader ng-show="isLoading" class="top-buffer-3"></mno-loader>
+        <div ng-repeat="request in requests  track by $index" ng-show="!isLoading">
```
alexnoox: Remove double space

src/app/components/impersonification-notification/impersonification-notification.html
```diff
@@ -0,0 +1,22 @@
...
+          <div ng-show="request.status == 'requested'">
```
alexnoox: Only requests with a status at `requested` should be fetched from the backend.

src/app/components/impersonification-notification/impersonification-notification.coffee
```diff
@@ -0,0 +1,49 @@
...
+  $scope.isLoading = true
+
+  $scope.initialize = (requests) ->
+    $scope.requests = requests
+    $scope.isLoading = false
+
+  $scope.requesterName = (request) ->
+    request.requester.name + ' ' + request.requester.surname
+
+  $scope.denyAccess = (index, request) ->
```
alexnoox: `request` is the functional object here, the arguments order should be switched.

src/app/components/impersonification-notification/impersonification-notification.coffee
```diff
@@ -0,0 +1,49 @@
...
+    MnoeUserAccessRequests.deny(request.id).then(
+      () ->
+        $scope.requests.splice(index, 1)
+        toastr.success('mno_enterprise.templates.impac.impersonification_notification.deny.success_toastr', {extraData: {name: $scope.requesterName(request)}})
+      (errors) ->
+        $log.error(errors)
+        toastr.error('mno_enterprise.templates.impac.impersonification_notification.deny.error_toastr', {extraData: {name: $scope.requesterName(request)}})
+    ).finally(-> $scope.isLoading = false)
+
+  $scope.approveAccess = (index, request) ->
```
alexnoox: `request` is the functional object here, the arguments order should be switched.

src/app/components/impersonification-notification/impersonification-notification.coffee
```diff
@@ -0,0 +1,49 @@
...
+
+  MnoeUserAccessRequests.list().then(
+    (requests) ->
+      $scope.initialize(requests)
+    (errors) ->
+      $log.error(errors)
+  )
+
+angular.module 'mnoEnterpriseAngular'
+  .directive('impersonificationNotification', ->
```
alexnoox: Could you transform this `directive` to a `component`?
And also use `controllerAs` instead of `$scope`?

src/app/components/impersonification-notification/impersonification-notification.html
x4d3: I will remove this condition.

src/app/components/impersonification-notification/impersonification-notification.coffee
```diff
@@ -1,49 +1,50 @@
...
+ImpersonificationNotificationCtrl = (toastr, MnoErrorsHandler, MnoeConfig, MnoeUserAccessRequests) ->
   'ngInject'
+  vm = this
+  vm.isEnabled = MnoeConfig.isImpersonationConsentRequired()
+  vm.isLoading = true
 
-  $scope.isLoading = true
+  vm.initialize = (result) ->
+    vm.requests = result
+    vm.isLoading = false
```
alexnoox: You could put this code directly in the promise `then()`

x4d3:
@alexnoox I've addressed your comment.
Regards,

alexnoox:
WHat I mean is that you can remove the `vm.initialize` function and move its content in the `then` directly:
```
MnoeUserAccessRequests.list().then(
    (result) ->
      vm.requests = result
      vm.isLoading = false
      (errors) ->
        MnoErrorsHandler.processServerError(errors)
    )
```

`vm.initialize` is not reused anywhere and it is just moving variables affectations at the other end of the file.

PR #293, opened 2017-08-24
Handle $ref in product JSON schemas

alachaum:
Summary of changes:
- Upgrade angular-schema-form to 1.0.0-alpha4
- Force multiples passes on JSON schemas to resolve $ref pointers


alachaum:
I'll wait for angular-schema-form to reach 1.0 and see if they support cyclic references. If they don't then we'll check $ref using a loop.

PR #295, opened 2017-08-27
Fix due tasks filter to exclude done tasks

xaun:
@alexnoox please review & merge


xaun:
@alexnoox please merge

PR #306, opened 2017-09-05
Public Pages - fix redirect logic

adamaziz15:
@ouranos @alexnoox I think this will make the redirects a bit cleaner. Now it doesn't go to the landing page at all if public pages are not enabled.


ouranos:
See mno-enterprise#444

PR #320, opened 2017-09-14
[MNO-686] Fix popover not closing

hedudelgado:


src/app/components/dashboard-apps-dock/no-apps-notification.html
```diff
@@ -1,4 +1,4 @@
-<div class="close top-right-corner" type="button" aria-label="Close" ng-click='toggle()'>x</div>
+<div class="close top-right-corner" type="button" aria-label="Close" ng-click='$parent.isPopoverShown = false'>x</div>
```
cesar-tonnoir: I think that should work, but @alexnoox would ask you to put that in a method :) Can you do this quick change plz?

hedudelgado: Hi @cesar-tonnoir , thans for your comment. Only one quesiton.
 I would be creating a controller only to handle that, Is really worth it to add "so much code" only for this?

alexnoox: It would be better to have a method `$parent.closePopover()`

hedudelgado: That's a better one! commit pushed @alexnoox

src/app/components/dashboard-apps-dock/dashboard-apps-dock.coffee
```diff
@@ -10,6 +10,7 @@ DashboardAppsDockCtrl = ($scope, $cookies, $uibModal, $window, $timeout, MnoeOrg
   $scope.popoverTemplateUrl = 'app/components/dashboard-apps-dock/no-apps-notification.html'
   $scope.isMarketplaceEnabled = MARKETPLACE_CONFIG.enabled
   $scope.isOnboardingEnabled = ONBOARDING_WIZARD_CONFIG.enabled
+  $scope.isPopoverShown = $scope.isOnboardingEnabled && $scope.isMarketplaceEnabled && !$scope.isLoading && _.isEmpty($scope.apps)
```
alexnoox: This should stay a function, `!$scope.isLoading` will only be evaluated at init time.

hedudelgado: Sorry @alexnoox , I went to add the fix blindly. In that case, can you please have a look at the new commit?

alexnoox:
@hedudelgado Please rebase on `1.1`

hedudelgado:
done @alexnoox 
Thanks

PR #321, opened 2017-09-14
Fix issue of modal being opened twice - Rotate refreh icon while loading

x4d3:


src/app/components/mnoe-tasks/mnoe-tasks.html
```diff
@@ -1,12 +1,12 @@
...
   <div class="flex-row">
-    <button class="btn btn-primary compose" ng-click="$ctrl.openCreateTaskModal()">
+    <button class="btn btn-primary compose" ng-click="$ctrl.openCreateTaskModal()" ng-disabled="$ctrl.isModalOpened">
       <span translate>mno_enterprise.templates.components.mnoe-tasks.compose_button_text</span>
       <i class="fa fa-pencil-square-o" aria-hidden="true"></i>
     </button>
     <mno-dropdown-filter-selector filters="$ctrl.tasksFilters" selected="$ctrl.selectedTasksFilter" on-select="$ctrl.onSelectFilter($event)"></mno-dropdown-filter-selector>
-    <div class="refresh-tasks" ng-click="$ctrl.onRefreshTasks()">
-      <i class="fa fa-refresh fa-2x fa-fw" aria-hidden="true"></i>
+    <div class="refresh-tasks" ng-click="$ctrl.onRefreshTasks()" ng-disabled="$ctrl.tasks.loading"  ng-class="{'disabled': $ctrl.tasks.loading}">
```
xaun: ng-disabled should be removed here

src/app/components/mnoe-tasks/mnoe-tasks.less
```diff
@@ -17,13 +17,22 @@ mnoe-tasks {
...
-        margin: 2px 2px 2px 7px;
+        padding: 5px;
+        margin: 2px 2px 2px 8px;
         color: @mno-tasks-column-sort-icon-active-color;
         &:hover {
           color: darken(@mno-tasks-column-sort-icon-active-color, 15%);
         }
+        &.disabled {
+          opacity: 0.5;
+          pointer-events: none;
```
xaun: I would not recommend using `pointer-events: none`, it does not give the user any feedback and may feel like the page is frozen.

I recommend using `pointer: progress` or `pointer: not-allowed`. Here are the available pointer style examples: https://www.w3schools.com/cssref/tryit.asp?filename=trycss_cursor.

PR #337, opened 2017-09-25
Fix subscription panel

hedudelgado:


src/app/views/provisioning/subscriptions.less
```diff
@@ -1,5 +1,10 @@
 #provisioning-subscriptions {
+ width: 90vw;
+ position: relative;
+ margin-left: -49vw;
+ left: 55%;
   .well {
     background-color: @provisioning-well-bg-color;
   }
+
```
alexnoox: Remove empty line

src/app/views/provisioning/subscriptions.less
```diff
@@ -1,5 +1,10 @@
 #provisioning-subscriptions {
+ width: 90vw;
+ position: relative;
+ margin-left: -49vw;
+ left: 55%;
```
alexnoox: I'm not sure about the 
```
margin-left: -49vw;
left: 55%;
```
It looks hacky to me. `width: 90vw;` is 90% of the viewport width, I think you can have a `margin-left` the size of the left side menu (I think there is a variable).

src/app/views/provisioning/subscriptions.less
```diff
@@ -1,5 +1,10 @@
 #provisioning-subscriptions {
+ width: 90vw;
```
alexnoox: `width: 90vw;` after a certain point will be too large Please add a `max-width` corresponding to the widest bootstrap breakpoint.

hedudelgado: Hey @alexnoox ,

I have tested in different screens, different zooms, and different browser sizes and seems to work fine. 
90vw will set the width to the 90% of the viewport, so this will be the minimum and maximum relative to the viewport. Am I wrong?

I have changed it anyway, but it would be great if you can get back to me on this.

src/app/views/provisioning/subscriptions.less
hedudelgado: There is no variable for the width of the sidebar menu. If there was a variable, using margin-left would not work due to its structure. 
This ID is wrapped within a container, which is wrapped within a container-fluid, so setting a margin-left would not work since it would be using the bootstrap container(ui-view) as a reference. I have used VW to be able to ignore its parent container and set margin-left and left is the only way to centre it I think.

I have created this codepen to be more clear on my explanation.
https://codepen.io/heDu/pen/mBRzMd

PR #340, opened 2017-09-26
[BUG] IE11 UI issues - widget refresh

adamaziz15:
@alexnoox This styling update should handle some UI bugs on IE11. ng-hide/ng-show has a lag on IE11 causing both elements to show simultaneously for a moment. This fixes that issue on IE11 and has no effect on chrome. (checked both just in case) Please review. :)

### DIFF
--- a/src/app/index.less
+++ b/src/app/index.less
@@ -72,6 +72,10 @@ body {
   z-index: 99999;
 }
 
+.content-template-wrapper.ng-hide {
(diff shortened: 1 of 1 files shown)


adamaziz15:
@alexnoox I did try applying it to all of them, but unfortunately there are some css classes coming from libraries that override the css if the class is not more specific than just `.ng-hide`

PR #347, opened 2017-09-27
Update isInteger for IE11

adamaziz15:
@alexnoox This PR should take care of some issues on the change password page. The function comes from ES6 which is not supported by IE11 so this replaces the function in that case. Please review.


adamaziz15:
@alexnoox Replaced with polyfill and tested locally, seems like everything is working properly 👍  Please review.

PR #354, opened 2017-10-09
Fix: Task is due "Today" is not appearing

x4d3:


src/app/components/mnoe-api/notifications.svc.coffee
```diff
@@ -83,13 +86,16 @@ angular.module 'mnoEnterpriseAngular'
           tls['mno_enterprise.templates.components.notifications.due.message.from'],
           tls['mno_enterprise.templates.components.notifications.due.message.details']
         ].join('</br>')
-
+        title = if isToday(task.due_date)
+          tls['mno_enterprise.templates.components.notifications.due.title_today']
```
alexnoox: We should instead use https://momentjs.com/docs/#/displaying/tonow/. It would display `in 2 days`, `in 1 day`, etc.

x4d3: I agree, but that was not part of the specs that wanted to display the date, except when it is "today"

PR #363, opened 2017-10-19
[MNOE-438] Fix install app button position

hedudelgado:


hedudelgado:
@ouranos

ouranos:
See #360

ouranos:
Close as we'll integrate feature/task into a release branch

PR #365, opened 2017-10-25
[MNOE-718] Theme Previewer

rheasunshine:
https://maestrano.atlassian.net/browse/MNOE-718

IE11 Bug fixes for theme-previewer:

IE11 does not support the color input type.  I have implemented a color picker that is compatible across browsers.

IE11 does not allow the use of css calculations with 3d transform, which was preventing the theme previewer panel from closing. I have changed the value from (100%-30px) to be 95%, leaving 5% of the panel visible.

### DIFF
--- a/src/app/components/theme-editor/theme-editor.html
+++ b/src/app/components/theme-editor/theme-editor.html
@@ -23,7 +23,7 @@
             <div class="form-group col-md-6" ng-repeat="(color, value) in theme">
               <label>{{labels[color]}} <br/><small class="text-muted">{{color}}</small></label>
               <spectrum-colorpicker
-                ng-model="theme[color]" ng-model-options="{ debounce: 300 }" ng-change="editor.update()" palette
+                ng-model="theme[color]" ng-model-options="{ debounce: 300 }" ng-change="editor.update()"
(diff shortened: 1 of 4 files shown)


src/app/components/theme-editor/theme-editor.html
```diff
@@ -17,16 +17,15 @@
...
         <uib-tab heading="Theme">
           <div class="row top-buffer-1">
             <div class="form-group col-md-6" ng-repeat="(color, value) in theme">
               <label>{{labels[color]}} <br/><small class="text-muted">{{color}}</small></label>
-              <input type="color" class="form-control"
-                     ng-model="theme[color]"
-                     ng-model-options="{ debounce: 300 }"
-                     ng-change="editor.update()">
+              <spectrum-colorpicker
+                ng-model="theme[color]" ng-model-options="{ debounce: 300 }" ng-change="editor.update()" palette
```
alexnoox: Are you sure you need the attribute `palette` here? Is it used to display it?

Reading the doc, it seems to me it is used to define the colors in the palette.

rheasunshine: It is unnecessary. I just pushed a cleanup

PR #366, opened 2017-10-26
 [MNOE-714] Check locales availability

ouranos:
Make sure that the locale is available before setting it.
Fix bug when a locale/I18n is disabled and the locale doesn't revert to
the default locale.

### DIFF
--- a/src/app/components/mno-locale-config/mno-locale-config.svc.coffee
+++ b/src/app/components/mno-locale-config/mno-locale-config.svc.coffee
@@ -21,6 +21,11 @@ angular.module 'mnoEnterpriseAngular'
       MnoeApiSvc.setBaseUrl(api_root)
       MnoeFullApiSvc.setBaseUrl(api_root)
 
+    # Check if the detected locale is available
+    filterAvailableLocales = (locale) ->
+      return unless locale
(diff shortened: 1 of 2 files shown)


ouranos:
Requires maestrano/mno-enterprise#549

src/app/components/mno-locale-config/mno-locale-config.svc.coffee
```diff
@@ -31,15 +36,15 @@ angular.module 'mnoEnterpriseAngular'
       found = path.match(re)
 
       # Ex found: ["/en/dashboard/", "en", index: 0, input: "/en/dashboard/"]
-      locale = found[1] if found?
+      locale = filterAvailableLocales(found[1]) if found?
 
       $q.resolve(locale)
```
fgourichon: What happens here if locale has not been found?

ouranos: It returns `undefined` so it will fallback to another locale.

See `@configure`:
```coffee
       $q.all(url: localeFromUrl(), user: localeFromUser()).then(
         (response) ->
           if l = response.url || response.user
             setLocale(l)
           else
             setFallbackStack($translate.use())
       )
```

PR #367, opened 2017-10-26
[MNOE-345] Fix display categories filter when only one

hedudelgado:
@alexnoox I added the calculation in the controller, but I would save this line of code and add it in the view. What do you think?


src/app/views/marketplace/marketplace.controller.coffee
```diff
@@ -59,6 +59,7 @@ angular.module 'mnoEnterpriseAngular'
           response = response.plain()
 
           vm.categories = response.categories
+          vm.displayCategories = vm.categories.length
```
alexnoox: This should be `vm.displayCategories = vm.categories.length > 1`

PR #368, opened 2017-10-29
[MNOE-750] Add missing locale for provisioning subscription

BrunoChauvet:


ouranos:
See maestrano/mno-enterprise#551

PR #371, opened 2017-11-02
[MNOE-750] Display SubscriptionEvent provisioning data

BrunoChauvet:
Feature preview:
screenshot from 2017-11-02 11-28-22

@alexnoox could you confirm that removing `json-formatter` and its configuration has no side effects? I haven't found pages where it was used previously.


alexnoox:
The json-formatter was only used on this page.

PR #380, opened 2017-11-15
[MNOE-788] Improve Devise configuration

ouranos:
* Add password expiration interceptor
* Configurable min password length

See maestrano/mno-enterprise#578


src/app/index.config.coffee
```diff
@@ -36,6 +36,22 @@ angular.module 'mnoEnterpriseAngular'
           $q.reject rejection
       }
 
+  .factory('PasswordExpiredInterceptor', ($log, $q,$location,$window) ->
+    return {
+      'responseError': (response) ->
+        if response.status == 403 && response.data.error && response.data.error == "Your password is expired. Please renew your password."
```
BrunoChauvet: Is it i18nable ?

src/app/components/mno-password/mno-password.coffee
```diff
@@ -7,19 +7,21 @@ angular.module 'mnoEnterpriseAngular'
...
         current: '=?',
+        minLength: '@',
         currentText: '@',
         newText: '@',
         confirmText: '@'
       },
       templateUrl: 'app/components/mno-password/mno-password.html',
       link: (scope, element, attrs) ->
         # Init args
+        scope.minLength = 6 if !scope.minLength
```
BrunoChauvet: Should the default min length values be aligned ? It is set to `6` here and `8` inside the `DEVISE_CONFIG`

src/app/index.config.coffee
ouranos: We actually match the response from the api which is not I18ned. The message we display to the user is I18ned

PR #386, opened 2017-11-22
[MNOE-793] Rename Admin Panel to Control Panel

enizor:


ouranos:
And https://github.com/maestrano/mno-enterprise-angular/blob/2.0/src/locales/en.json#L938

ouranos:
All good, could you squash the commits together? Thanks

enizor:
rebased

PR #387, opened 2017-11-23
[MNOE-710] Rename role: Super Admin -> Admin, Admin -> Manager

enizor:


src/locales/en.json
```diff
@@ -617,12 +617,12 @@
   "mno_enterprise.templates.dashboard.organization.members.role": "Role",
   "mno_enterprise.templates.dashboard.organization.members.roles.member": "Member",
   "mno_enterprise.templates.dashboard.organization.members.roles.power_user": "Power User",
```
ouranos: Power User has been removed in  #276
Check that the locales is no longer used an remove it. Same for "Invited (Power User)"

PR #390, opened 2017-11-29
[MNOE-710] Roles update - fix forgotten ones

enizor:


enizor:
Will also remove all code related to power users

enizor:
Power Users code should remain - deleting commit

PR #395, opened 2017-12-05
[MNOE-561] Update company name on refresh

manu-d:
- Company was fetched from MnoeCurrentUser, which relied on the cache from mno-enterprise. Therefore, the company name was not updated on refresh.

- MnoeOrganizations.selected is set right after MnoeOrganizations.selectedId is set.
Since the watch that triggers selectOrganization() is on MnoeOrganizations.selectedId, MnoeOrganizations.selected can be nil at this point, so we keep what we had as a fallback

### DIFF
--- a/src/app/components/dashboard-company-selectbox/dashboard-company-selectbox.coffee
+++ b/src/app/components/dashboard-company-selectbox/dashboard-company-selectbox.coffee
@@ -27,7 +27,7 @@ DashboardCompanySelectboxCtrl = ($scope, $state, $stateParams, $uibModal, MnoeCu
     selectBox.close()
 
   selectOrganization = ->
-    selectBox.organization = _.find(MnoeCurrentUser.user.organizations, { id: MnoeOrganizations.selectedId })
+    selectBox.organization = MnoeOrganizations.getSelected()?.organization || _.find(MnoeCurrentUser.user.organizations, { id: MnoeOrganizations.selectedId })
 
   # Toggle the select box
   selectBox.toggle = ->


src/app/components/dashboard-company-selectbox/dashboard-company-selectbox.coffee
```diff
@@ -27,7 +27,7 @@ DashboardCompanySelectboxCtrl = ($scope, $state, $stateParams, $uibModal, MnoeCu
     selectBox.close()
 
   selectOrganization = ->
-    selectBox.organization = _.find(MnoeCurrentUser.user.organizations, { id: MnoeOrganizations.selectedId })
+    selectBox.organization = MnoeOrganizations.selected.organization || _.find(MnoeCurrentUser.user.organizations, { id: MnoeOrganizations.selectedId })
```
xaun: Maybe use `MnoeOrganizations.getSelected()` as it was designed as a getter. `MnoeOrganizations.selected` should really be `_selected` to hint that it's designed as a private property. This means if in the future some changes a made to the way the `MnoeOrganization` service retrieves a stored selected organisation, this code can go unchanged.

src/app/components/dashboard-company-selectbox/dashboard-company-selectbox.coffee
```diff
@@ -27,7 +27,7 @@ DashboardCompanySelectboxCtrl = ($scope, $state, $stateParams, $uibModal, MnoeCu
     selectBox.close()
 
   selectOrganization = ->
-    selectBox.organization = _.find(MnoeCurrentUser.user.organizations, { id: MnoeOrganizations.selectedId })
+    selectBox.organization = MnoeOrganizations.getSelected().organization || _.find(MnoeCurrentUser.user.organizations, { id: MnoeOrganizations.selectedId })
```
alexnoox: I think it would be safer to use a navigation operator here:
`MnoeOrganizations.getSelected()?.organization `

PR #413, opened 2018-01-08
[MNOE-636] Error message when no subscription available for org currency

manu-d:


src/app/views/provisioning/order.html
```diff
@@ -16,6 +16,12 @@ <h1 translate>mno_enterprise.templates.dashboard.provisioning.order.title</h1>
           </div>
         </div>
 
+        <div class="row td-row" ng-if="vm.subscription.product.product_pricings.length === 0">
```
xaun: You could just go: `ng-if="!vm.subscription.product.product_pricings.length"` as `0` is falsey in JS.

manu-d: Keep consistency with https://github.com/maestrano/mno-enterprise-angular/blob/2.0/src/app/views/provisioning/subscriptions.html#L30

src/app/views/marketplace/marketplace-product.controller.coffee
```diff
@@ -30,6 +30,9 @@ angular.module 'mnoEnterpriseAngular'
           $state.go('home.marketplace') unless vm.product?
       ).finally(-> vm.isLoading = false)
 
+    vm.planAvailableForCurrency = (plan) ->
+      _.includes(_.map(plan.prices, 'currency'), vm.orgCurrency)
```
ouranos: You could also do:
```ruby
_.find(plan.prices, { 'currency': vm.orgCurrency })
```

PR #414, opened 2018-01-08
[MNOE-470] Currency Selection (Customer / Tenant)

MAhsenArif:


manu-d:
Same as https://github.com/maestrano/mno-enterprise/pull/629, can you leave only the english translation?

src/app/views/company/settings/organization-settings.directive.coffee
```diff
@@ -52,6 +53,16 @@ DashboardOrganizationSettingsCtrl = ($scope, $window, MnoeOrganizations, Utiliti
     f = $scope.forms
     $scope.isChanged() && f.settings.$valid
 
+  $scope.isCurrencyChangeShown = ->
+    if MnoeOrganizations.role.isSuperAdmin()
+      $scope.availableBillingCurrencies()
```
ouranos: What the advantage of this vs:
```coffescript
$scope.isCurrencyChangeShown = -> MnoeOrganizations.role.isSuperAdmin()
$scope.currencies = MnoeConfig.availableBillingCurrencies()
```
?

src/app/views/company/settings/organization-settings.html
```diff
@@ -19,6 +19,15 @@
         <input type="text" class="form-control" ng-model="model.name" required/>
       </div>
     </div>
+
+    <div class="form-group" ng-if="isCurrencyChangeShown()">
+      <label class="col-sm-3 control-label">{{ 'mno_enterprise.templates.dashboard.organization.settings.billing_currency' | translate }}</label>
+      <div class="col-sm-6">
+        <select ng-model="model.billing_currency" class="form-control">
+          <option ng-repeat="currency in currencies" value="{{currency}}">{{currency}}</option>
```
ouranos: You can use one-way binding here (`{{::currency}`) as the list is static

src/app/views/company/settings/organization-settings.directive.coffee
MAhsenArif: I added it like this as a form of secondary validation check to never load currencies if the user is not Super Admin.

ouranos: Ok fair enough. Maybe rename it to `$scope.getAvailableBillingCurrencies()` to make it more explicit?

MAhsenArif:
@ouranos Feedback incorporated.

PR #415, opened 2018-01-10
[MNOE-872] Session timeout modal

enizor:
An interceptor will check all the sent request and set / reset a timer.
WIP implementation.

What still needs to be done:

- [x] i18n
- [x] Do not reset timer on login/logout
- [x] Drop MnoConfirm for a custom modal
- [x] Have a countdown on the modal to hide the 'stay logged in' button
  when the user has actually timed out
- [x] Get the timeout setting from MNOE


enizor:
@ouranos for the review

MAhsenArif:
This can be closed as task is continued on https://github.com/maestrano/mno-enterprise-angular/pull/432

PR #418, opened 2018-01-16
[MNOE-894] Logo redirects to the dashboard

enizor:
@ouranos


enizor:
It will work in this case too, the state `home.impac` displays the apps dock / impac dashboard depending on the user and organization

PR #421, opened 2018-01-30
Add v2 Bolt configuration

MarcoBgn:
@cesar-tonnoir this is minimal, added to track the feature


cesar-tonnoir:
@MarcoBgn this should not be added to the framework otherwise your bolt would be activated by default on all the frontends. I'll do the similar change directly on _Kappa_.

PR #425, opened 2018-02-01
[MNOE-406] Added logic to request for unscoped data when showing dashboard apps

MAhsenArif:


src/app/components/mnoe-api/app-instances.svc.coffee
```diff
@@ -6,7 +6,7 @@ angular.module 'mnoEnterpriseAngular'
     @appInstances = []
 
     appInstancesPromise = null
-    @getAppInstances = ->
+    @getAppInstances = (force = false) ->
```
ouranos: The `force` is here to force unscoping?
In this case, it might more obvious to call it `unscoped`? And also add a quick comment on both `@getAppInstance` and `@fetchAppInstance`

src/app/components/mnoe-api/app-instances.svc.coffee
```diff
@@ -33,14 +35,16 @@ angular.module 'mnoEnterpriseAngular'
...
+    fetchAppInstances = (unscoped) ->
       # Workaround as the API is not standard (return a hash map not an array)
       # (Prefix operation by '/' to avoid data extraction)
       # TODO: Standard API
       defer = $q.defer()
       MnoeOrganizations.get().then(
         ->
-          _self.appInstancesPromise = MnoeApiSvc.one('organizations', MnoeOrganizations.selectedId).one('/app_instances').get().then(
+          query = if unscoped == true then '?unscoped=true' else ''
+          _self.appInstancesPromise = MnoeApiSvc.one('organizations', MnoeOrganizations.selectedId).one('/app_instances' + query).get().then(
```
ouranos: For query parameters, [you can pass an object to `get()`](https://github.com/mgonto/restangular#element-methods):
```coffeescript
params = if unscoped then {unscoped: true} else {}
xxx.one('/app_instances').get(params)
```

PR #432, opened 2018-02-22
[MNOE-872] Session timeout modal 

MAhsenArif:
Continued https://github.com/maestrano/mno-enterprise-angular/pull/415 in this PR


src/app/components/mno-session-timeout/mno-session-timeout.svc.coffee
```diff
@@ -26,21 +26,31 @@ angular.module 'mnoEnterpriseAngular'
 
       $scope.countdown = 10
 
-      $interval((-> $scope.countdown -= 1), 1000, 10)
+      $interval((
+        ->
+          $scope.countdown -= 1
+          if $scope.countdown == 0
```
ouranos: You can have a one liner `$scope.logOff(true) if $scope.countdown == 0`

src/app/index.config.coffee
```diff
@@ -69,12 +69,13 @@ angular.module 'mnoEnterpriseAngular'
...
 
   .config ($httpProvider) ->
-    $httpProvider.interceptors.push ($q, $injector, DEVISE_CONFIG) ->
+    $httpProvider.interceptors.push ($q, $injector, DEVISE_CONFIG, $rootScope) ->
       {
         request: (config) ->
           # Intercept requests made to the API and reset the sessions timeout
           if DEVISE_CONFIG.timeout_in > 0
-            if config.url.includes('mnoe') && !config.url.match(/(sign_out|sign_in|app_instances_sync)/)
+            if config.url.includes('mnoe') && !config.url.match(/(sign_out|sign_in|app_instances_sync)/) && !$rootScope.timeoutSet
```
ouranos: If I understand properly you're only resetting the timeout if it's not set.
I don't think that's correct as subsequent API requests are updating devise `last_active_at` so should also reset the frontend timer.

See:
* [AppInstancesSyncController](https://github.com/maestrano/mno-enterprise/blob/3.3/api/app/controllers/mno_enterprise/jpi/v1/app_instances_sync_controller.rb#L5)
* [MnoEnterprise::ApplicationController](https://github.com/maestrano/mno-enterprise/blob/3.3/core/app/controllers/mno_enterprise/application_controller.rb#L51)

MAhsenArif: Removed the flag as it was indeed preventing requests from resetting the timer every time an API request was received.

src/app/index.config.coffee
```diff
@@ -68,6 +68,18 @@ angular.module 'mnoEnterpriseAngular'
               $q.reject rejection
       }
 
+  .config ($httpProvider) ->
+    $httpProvider.interceptors.push ($q, $injector, DEVISE_CONFIG, $rootScope) ->
```
ouranos: `$rootScope` dependency is no longer needed

src/locales/en.locale.json
```diff
@@ -733,6 +733,11 @@
   "mno_enterprise.bundle.app_selection.cancel": "Cancel",
   "mno_enterprise.bundle.what_is_maestrano.title": "Powerful cloud technology built for small to medium business",
   "mno_enterprise.bundle.what_is_maestrano.description": "Maestrano synchronizes data from your favourite business apps, making everything accessible on one platform.",
+  "mno_enterprise.auth.sessions.timeout.header": "Session Timeout",
```
ouranos: These modals need to be added on to mnoe as well. I'll do it

PR #420, opened 2018-01-18
[MNOE-890] Reseller's products on landing page

enizor:

PR #435, opened 2018-02-27
[MNOE-890] Reseller's products on landing page

MAhsenArif:
See #420


MAhsenArif:
@ouranos The ternary condition `a ? b : c` throws syntax error in coffee script. I've reverted to using the `if a then b else c` syntax as that is the recommended one for coffee.

PR #439, opened 2018-03-05
[MNOE-845] Update locale when Admin Panel default language changes

MAhsenArif:


MAhsenArif:
@ouranos `$translate.use()` is picking up the language to be `en`. The [preferred language](https://github.com/maestrano/mno-enterprise-angular/blob/2.0/src/app/index.config.coffee#L67) is being based on browser language.

PR #448, opened 2018-03-27
[MNOE-1000] product usage based pricing

iseessel:


adamaziz15:
lgtm

src/app/components/mno-local-products/mno-local-product.coffee
```diff
@@ -17,6 +17,9 @@ angular.module 'mnoEnterpriseAngular'
       org = _.find(user.organizations, { id: currentOrg.id })
       MnoeOrganizations.role.atLeastAdmin(org.current_user_role)
 
+    vm.noPricingFound = (plan) ->
```
ouranos: Too much negation here, it's more understandable if you use affirmative form.
Also the method name is slightly misleading (a price can exist but the page is public)

PR #457, opened 2018-04-10
[MNOE-1032] Fix incorrect role UI

iseessel:


iseessel:
@ouranos I had done that initially, but the logic is a bit different than the other place it is used, but agreed.

PR #467, opened 2018-04-23
[OPAL-305] Update Product pricings when organization is changed

MAhsenArif:
- ~~When fetching apps/products, we do not get updated pricing details since we are showing cached versions of the apps against that organization.~~

- Allow Product pricings to be updated when organization is changed


ouranos:
Rebased to discard reverted commits :)

PR #471, opened 2018-04-26
[MNOE-951] Fix locale reset on hard refresh

adamaziz15:
This should fix the locale refresh bug. Calling `$translate.refresh()` updates the translations table being used. Otherwise when the fallback stack is set the changes wont take effect until you switch tabs or refresh again. 

I also noticed a bug as the loader wouldn't display in the chosen language. This was related to the same problem but the fix didn't work on the loader text as it was using just the translate directive (which doesnt set a watcher). By using the translate filter it sets a watcher that causes the text to change when `$translate.refresh()` is called.


src/app/index.config.coffee
```diff
@@ -59,10 +59,11 @@ angular.module 'mnoEnterpriseAngular'
     }
 
     availableLocales = _.map(I18N_CONFIG.available_locales, 'id')
+    preferredLocale = I18N_CONFIG.preferred_locale.slice(0,2)
```
ouranos: What is the problem when using the full code?

ouranos: @adamaziz15 ping

adamaziz15: @ouranos Ah, angular-translate does not support the full codes by default. See: https://angular-translate.github.io/docs/#/guide/09_language-negotiation

PR #485, opened 2018-05-14
[MNOE-1083] Company Cart System

MAhsenArif:


src/app/components/mnoe-api/provisioning.svc.coffee
```diff
@@ -82,32 +87,37 @@ angular.module 'mnoEnterpriseAngular'
...
       deferred = $q.defer()
       MnoeOrganizations.get().then(
         (response) ->
-          subscriptionsApi(response.organization.id).getList().then(
+          subscriptionsApi(response.organization.id).getList(params).then(
             (response) ->
               deferred.resolve(response)
           )
       )
+      _self.subscriptionsPromise = deferred.promise if cart
```
adamaziz15: Since this is only used in the cart, maybe we should rename it `cartSubscriptionsPromise`

src/app/views/provisioning/subscriptions.controller.coffee
```diff
@@ -3,39 +3,95 @@ angular.module 'mnoEnterpriseAngular'
...
     vm.cancelSubscription = (subscription, i) ->
+      if vm.cartSubscriptions
+        subscription.cart_entry = true
+        headerText = 'mno_enterprise.templates.dashboard.provisioning.subscriptions.cart.cancel_modal.title'
+        bodyText = 'mno_enterprise.templates.dashboard.provisioning.subscriptions.cart.cancel_modal.body'
+      else
+        headerText = 'mno_enterprise.templates.dashboard.provisioning.subscriptions.cancel_modal.title'
+        bodyText = 'mno_enterprise.templates.dashboard.provisioning.subscriptions.cancel_modal.body'
+
+      # TODO: When cart item is removed, we want to update the cart subscriptions
```
adamaziz15: Can we remove this todo? Looks like this is handled on line 33

src/locales/en.json
```diff
@@ -760,8 +769,12 @@
   "mno_enterprise.templates.dashboard.provisioning.subscriptions.new": "New install",
   "mno_enterprise.templates.dashboard.provisioning.subscriptions.change": "Change plan",
   "mno_enterprise.templates.dashboard.provisioning.subscriptions.modify": "Modify",
+  "mno_enterprise.templates.dashboard.provisioning.subscriptions.cart.cancel": "Remove Item",
```
adamaziz15: Can this translation be removed? It looks like it isn't being used.

PR #489, opened 2018-05-18
[OPAL-433] Add phone to address input

iseessel:


src/app/components/mno-address-input/mno-address-input.html
```diff
@@ -31,4 +31,10 @@
       </select>
     </div>
   </div>
+  <div class="form-group">
+    <div class="col-sm-6 address-field">
+      <label class="address-sub-label">{{ 'mno_enterprise.templates.components.mno_address.phone' | translate }}</label>
+      <input type="text" class="form-control" ng-model="address.phone"/>
+    </div>
+  </div>
 </div>
```
adamaziz15: Phone will just be a text input right? I think this could possibly cause some confusion for the user when it comes to country code. Should we have a `phone_country_code` attr on the address model in mnohub? I'm thinking we should have it set-up just like the User model in MnoHub.

iseessel: @adamaziz15 @ouranos  I can do that and its probably better that way. 

Just not sure how urgent this ticket is to be released, but would take me ~4h to change.

PR #493, opened 2018-05-22
[OPAL-1116] Display Custom Data

paulstuartparker:


src/app/views/provisioning/subscription.controller.coffee
```diff
@@ -4,7 +4,7 @@ angular.module 'mnoEnterpriseAngular'
     vm = this
 
     vm.isLoading = true
-
+    vm.customDataPresent = false
```
iseessel: no

src/app/views/provisioning/subscription.controller.coffee
```diff
@@ -19,12 +18,14 @@ angular.module 'mnoEnterpriseAngular'
...
-          MnoeMarketplace.findProduct(id: vm.subscription.product_id).then(
-            (response) ->
-              vm.schema = if response.custom_schema then JSON.parse(response.custom_schema) else {}
-              vm.form = if response.asf_options then JSON.parse(response.asf_options) else ["*"]
+
+          MnoeMarketplace.fetchCustomSchema(vm.subscription.product_id).then((response) ->
+            # Some products have custom schemas, whereas others do not.
+            resp = JSON.parse(response)
+            vm.schema = if resp?.json_schema then resp.json_schema else resp
+            vm.form = if resp?.asf_options then resp.asf_options else ["*"]
```
adamaziz15: can these be refactored this way?
```
vm.schema = resp?.json_schema || resp
vm.form = resp?.asf_options || ["*"]
```

src/app/views/provisioning/subscription.html
```diff
@@ -92,8 +92,8 @@ <h3 class="text-color" translate>mno_enterprise.templates.dashboard.provisioning
       <div class="col-sm-6">
         <div class="well">
           <h3 class="text-color" translate>mno_enterprise.templates.dashboard.provisioning.subscription.custom_data</h3>
-          <span ng-show="!vm.subscription.custom_data" translate>mno_enterprise.templates.dashboard.provisioning.subscription.no_custom_data</span>
-          <form name="vm.detailsForm" ng-show="vm.subscription.custom_data" sf-schema="vm.schema" sf-form="vm.form" sf-model="vm.model" sf-options="{ formDefaults: { readonly: { true } }}"></form>
+          <span ng-show="!vm.schema" translate>mno_enterprise.templates.dashboard.provisioning.subscription.no_custom_data</span>
```
adamaziz15: Would probably be better to have `ng-hide="vm.schema"` here

src/app/views/provisioning/subscription.controller.coffee
paulstuartparker: It is more idiomatic in coffeescript to use an if-then-else one liner: https://coffeescript.org/#conditionals

PR #495, opened 2018-05-23
[MNOE-1131] [MNOE-1133] Fix Provisioning Bugs

paulstuartparker:


src/app/views/provisioning/order.controller.coffee
```diff
@@ -52,7 +53,7 @@ angular.module 'mnoEnterpriseAngular'
           # Filters the pricing plans not containing current currency
           vm.filterCurrencies()
           MnoeProvisioning.setSubscription(vm.subscription)
-        ).finally(-> vm.isLoading = false)
+        )
```
iseessel: Good -- there is a vm.isLoading = false below as well.

src/app/views/provisioning/order.controller.coffee
```diff
@@ -29,7 +29,7 @@ angular.module 'mnoEnterpriseAngular'
       vm.filteredPricingPlans = _.filter(vm.subscription.product.pricing_plans,
         (pp) -> !vm.pricedPlan(pp) || _.some(pp.prices, (p) -> p.currency == vm.orgCurrency)
       )
-
+    # filter currencies if we are using a cached subscription
```
iseessel: Comment belongs below.

src/app/views/provisioning/order.controller.coffee
```diff
@@ -72,6 +71,13 @@ angular.module 'mnoEnterpriseAngular'
         )
         .finally(() -> vm.isLoading = false)
     else
+      currenciesArray = []
+      _.forEach(vm.subscription.product.pricing_plans,
+        (pp) -> _.forEach(pp.prices, (p) -> currenciesArray.push(p.currency)))
+
+      vm.currencies = _.uniq(currenciesArray)
+      vm.filteredPricingPlans = if vm.subscription?.product?.pricing_plans then vm.filterCurrencies() else []
```
iseessel: we don't need the conditional anymore because we've already checked if we have a cached subscription.

src/app/views/provisioning/order.controller.coffee
```diff
@@ -72,6 +71,13 @@ angular.module 'mnoEnterpriseAngular'
         )
         .finally(() -> vm.isLoading = false)
     else
+      currenciesArray = []
+      _.forEach(vm.subscription.product.pricing_plans,
```
iseessel: maybe abstract out 74-78 in a function to avoid duplication above.

src/app/views/provisioning/confirm.html
```diff
@@ -94,7 +94,7 @@ <h1 translate>mno_enterprise.templates.dashboard.provisioning.confirm.title</h1>
     </div>
 
     <span class="pull-right top-buffer-1">
-      <button class="btn btn-primary" ng-show="vm.orderEditable()" ng-click="vm.editOrder()" translate>mno_enterprise.templates.dashboard.provisioning.confirm.edit_order</button>
+      <button class="btn btn-primary" ng-show="vm.orderEditable()" ng-click="vm.editOrder(false)" ng-disabled="vm.disableEdit" translate>mno_enterprise.templates.dashboard.provisioning.confirm.edit_order</button>
```
iseessel: no longer using the vm.disableEdit

src/app/views/provisioning/confirm.controller.coffee
```diff
@@ -60,12 +60,14 @@ angular.module 'mnoEnterpriseAngular'
       ).finally(-> vm.isLoading = false)
 
     vm.orderEditable = () ->
+      return true if vm.subscription.product?.custom_schema
```
iseessel: Can we move the custom schema comment and add a comment for the skipPricingPlans.

src/app/views/provisioning/order.controller.coffee
```diff
@@ -26,6 +34,18 @@ angular.module 'mnoEnterpriseAngular'
           vm.subscription = response.subscription
         )
 
+    vm.filterCurrencies = () ->
+      vm.filteredPricingPlans = _.filter(vm.subscription.product.pricing_plans,
+        (pp) -> !vm.pricedPlan(pp) || _.some(pp.prices, (p) -> p.currency == vm.orgCurrency)
```
ouranos: Please reuse `ProvisioningHelper.planForCurrency`

src/app/views/provisioning/order.controller.coffee
```diff
@@ -43,27 +43,24 @@ angular.module 'mnoEnterpriseAngular'
...
       MnoeMarketplace.getProduct(vm.productId).then(
         (response) ->
           vm.subscription.product = response
-
           # Get all the possible currencies
           populateCurrencies()
-
-          # Set a default currency
           selectDefaultCurrency()
+
```
iseessel: extra line

src/app/services/provisioning-helper/provisioning-helper.coffee
```diff
@@ -10,4 +10,7 @@ angular.module 'mnoEnterpriseAngular'
         (pp) -> !_self.pricedPlan(pp) || _.some(pp.prices, (p) -> p.currency == currency)
       )
 
+    @skipPricingPlans = (product) ->
+      product.product_type == 'application' && (!product.single_billing_enabled || !product.billed_locally)
```
ouranos: @paulstuartparker @iseessel @adamaziz15 @yuritomanek (sorry for the spam)

After discussion with @BrunoChauvet, I think we shouldn't not skip the pricing plan step when then product is billed externally. We should only skip it when the app doesn't support single billing.

See OPALSUP-187

BrunoChauvet: You should actually skip the pricing selection if Single Billing is disabled AND Provisioning is disabled
We could face the scenario where an application is provisioned externally but and the billing is managed by the application. In this case we still need to ask the user to select a pricing plan to pass down to the external system.

ouranos: It's the `Product#externally_provisioned` flag. In this case, even if there's no single billing we still need to tell the external application which pricing plan they need to provision.

New behaviour, skip pricing selection for *applications* when:
* `product.single_billing_enabled == false` AND `product.externally_provisioned` == false

It might be easier to redefine it this way to avoid double negation:
```coffeescript
@skipPricingPlans = (product) ->
  product.product_type == 'application' && !product.single_billing_enabled && !product.externally_provisioned

# Or if you want to avoid too much negation
@displayPricingPlans = (product) ->
  product.product_type != 'application' || product.single_billing_enabled || product.externally_provisioned

```

@yuritomanek  @BrunoChauvet Do you agree?

PR #497, opened 2018-05-25
[MNOE-1139] AppManagement: Subscriptions Events should show in Order History Tab

MAhsenArif:


src/app/views/apps-management/app-management/app-management.html
```diff
@@ -227,35 +227,46 @@ <h3>
...
-                  <div class='col-sm-4'>
-                    <label>
-                      {{ 'mno_enterprise.templates.dashboard.app_management.manage.tab.order_history.offer_name' | translate }}
-                    </label>
-                  </div>
-                  <div class='col-sm-4'>
-                    <label>
-                      {{ 'mno_enterprise.templates.dashboard.app_management.manage.tab.order_history.status' | translate }}
-                    </label>
+              <div ng-if="!vm.subscriptionsHistory">
```
iseessel: Can we change this to ```ng-show="_.isEmpty(vm.subscriptionsHistory)"``` because I think the response will be an empty array, not false.

src/app/views/apps-management/app-management/app-management.html
```diff
@@ -227,35 +227,46 @@ <h3>
...
+                {{ 'mno_enterprise.templates.dashboard.app_management.manage.tab.order_history.no_history' | translate }}
+              </div>
+              <div ng-if="vm.subscriptionsHistory">
```
iseessel: Can we change to ```ng-hide="_.isEmpty(vm.subscriptionsHistory)"``` because I think it will instantiate to an empty array not false.

src/app/views/apps-management/app-management/app-management.html
MAhsenArif: @iseessel Actually this can be `nil`/`false` since this [line](https://github.com/maestrano/mno-enterprise-angular/pull/497/files#diff-90d18dde8e61e1c7adb35facf925e5e4R111) may never get triggered if there is no current subscription, thus never setting the `vm.subscriptionsHistory` variable.

src/app/views/apps-management/app-management/app-management.html
MAhsenArif: Same reasoning as above.

src/app/views/apps-management/app-management/app-management.html
iseessel: @MAhsenArif Ah, the use case I was thinking about is when the subscription is available, but there are no subscription events -- but come to think about it that will never happen.

PR #498, opened 2018-05-25
[MNOE-1140] Can provision app from admin panel but not marketplace

NickenChuggets:


src/app/components/mno-app-install-btn/mno-app-install-btn.coffee
```diff
@@ -167,7 +167,7 @@ angular.module 'mnoEnterpriseAngular'
             products = response.products?.products
             currency = MnoeOrganizations.selected.organization.billing_currency || MnoeConfig.marketplaceCurrency()
             plans = vm.app.pricing_plans
-            if !plans[currency]
+            if !plans[currency] && plans.default[0]?.pricing_type != "free"
```
ouranos: Shouldn't you use `ProvisioningHelper.pricedPlan`?

iseessel:
@ouranos @adamaziz15 @NickenChuggets What about when the pricing plan is managed by the application? Tested these changes -- and I'm not able to provision mail chimp from the dashboard, but I am able to provision from admin panel.

src/app/components/mno-app-install-btn/mno-app-install-btn.coffee
```diff
@@ -167,8 +167,7 @@ angular.module 'mnoEnterpriseAngular'
             products = response.products?.products
             currency = MnoeOrganizations.selected.organization.billing_currency || MnoeConfig.marketplaceCurrency()
             plans = vm.app.pricing_plans
-            if !plans[currency]
-              vm.arePlansAvailable = false
+            vm.arePlansAvailable = !plans[currency] || !ProvisioningHelper.pricedPlan(plans.default[0])
```
iseessel: Seems like this logic is reversed.

`vm.arePlansAvailable = plans[currency] || ...`

i.e. plans are available if the plans have our currency.

Also why are we only checking the first default plan -- there are products with multiple default plans.

iseessel: We also need to account for times where product pricings are skipped for products.

src/app/services/provisioning-helper/provisioning-helper.coffee
```diff
@@ -5,6 +5,14 @@ angular.module 'mnoEnterpriseAngular'
     @pricedPlan = (plan) ->
       plan.pricing_type not in PRICING_TYPES['unpriced']
 
+    @containsPricedPlan = (plans) ->
+      res = false
+      plans?.forEach( (plan) ->
```
ouranos: Use [some](https://lodash.com/docs/4.17.10#some)

src/app/components/mno-app-install-btn/mno-app-install-btn.coffee
ouranos:
```coffeescript
# TODO: Comments
vm.orderingPossible = plans[currency] || skipPricing || !ProvisioningHelper.hasPricedPlan(plans.default)
```

src/app/components/mno-app-install-btn/mno-app-install-btn.coffee
```diff
@@ -202,6 +200,8 @@ angular.module 'mnoEnterpriseAngular'
             # Is the product externally provisioned
             vm.isExternallyProvisioned = vm.isProvisioningEnabled && (product?.product_type == 'application' || product?.externally_provisioned)
 
+            vm.orderPossible = plans[currency] || !ProvisioningHelper.hasPricedPlan(plans.default) || ProvisioningHelper.skipPriceSelection(product)
+
```
adamaziz15: I do see an issue with this logic. Lets say I have an app that has default plans that include a priced plan as well as a usage based plan. The call to 
 `ProvisioningHelper.hasPricedPlan(plans.default)` will return true even though there is a plan that is not priced. I actually think that `vm.orderPossible` should be true for the following conditions:
```
vm.orderPossible = plans[currency] || plans.default[0] || ProvisioningHelper.skipPriceSelection(product)
```
This way as long as there is a default plan the button will be enabled. It will say no plans available in your currency but once you click provision you can select from what is available.

src/app/services/provisioning-helper/provisioning-helper.coffee
```diff
@@ -5,6 +5,9 @@ angular.module 'mnoEnterpriseAngular'
     @pricedPlan = (plan) ->
       plan.pricing_type not in PRICING_TYPES['unpriced']
 
+    @hasPricedPlan = (plans) ->
```
iseessel: Are we using this anywhere?

src/app/components/mno-apps/mno-app.coffee
```diff
@@ -38,13 +38,24 @@ angular.module 'mnoEnterpriseAngular'
 
     vm.pricedPlan = ProvisioningHelper.pricedPlan
 
-    getPricingPlans = () ->
+    # Ordering is possible if
```
iseessel: Consider refactoring? 

```
    orderPossible = (plans, product) ->
      if product
        vm.orderPossible = plans.default[0] || plans[currency] || ProvisioningHelper.skipPriceSelection(product)
      else
        vm.orderPossible = plans.default[0] || plans[currency]

    getPricingPlans = (product) ->
      if vm.isPriceShown
        plans = vm.app.pricing_plans
        vm.pricing_plans = plans[vm.currency]
        orderPossible(plans, product)
```

adamaziz15: For `orderPossible` I don't think the `plans[currency]` part is necessary. If any plans exist at all then there will always be a `plans.default[0]`.

For `getPricingPlans` it's probably ok as is. It's more explicit with the if statement as it makes it clear that we're checking if the order is possible *because* there was no `plans[vm.currency]`

src/app/services/provisioning-helper/provisioning-helper.coffee
adamaziz15: @NickenChuggets PR looks good once this is removed 👍

src/app/components/mno-apps/mno-app.html
```diff
@@ -109,10 +109,10 @@ <h2 class="media-heading">{{vm.app.name}}</h2>
       <div class="col-md-12">
         <div class="well">
           <h2>{{ 'mno_enterprise.templates.dashboard.marketplace.show.price' | translate }}</h2>
-          <div translate ng-if="!vm.arePlansAvailable">
+          <div translate ng-if="!vm.orderPossible">
```
adamaziz15: This should be `ng-if="!vm.pricing_plans"`

If there are no plans in your currency, but the order is still possible, we still want the "No plans available in your currency" message to appear. Otherwise there is just an empty pricing section.

src/app/components/mno-apps/mno-app.html
```diff
@@ -109,10 +109,10 @@ <h2 class="media-heading">{{vm.app.name}}</h2>
...
             {{ 'mno_enterprise.templates.dashboard.marketplace.show.no_pricing_plans_found' | translate }}
           </div>
-          <ul ng-if="vm.arePlansAvailable">
+          <ul ng-if="vm.orderPossible">
```
adamaziz15: Should be `ng-if="vm.pricing_plans"` so that this section is hidden if there are no plans.

src/app/components/mno-apps/mno-app.coffee
```diff
@@ -38,13 +38,24 @@ angular.module 'mnoEnterpriseAngular'
...
+    # -Plans exist in the current currency
+    # -The app only has free or pay as you go plans
+    # -Price selection can be skipped
+    orderPossible = (plans, product) ->
+      if product
+        vm.orderPossible = plans.default[0] || ProvisioningHelper.skipPriceSelection(product)
+      else
+        vm.orderPossible = plans.default[0]
+
+    getPricingPlans = (product) ->
```
adamaziz15: #### For this `getPricingPlans` function:
```coffeescript
getPricingPlans = (product) ->
  if vm.isPriceShown
    plans = vm.app.pricing_plans
    plans_for_currency = ProvisioningHelper.planForCurrency(plans, vm.currency)
    if _.isEmpty(plans_for_currency)
      orderPossible(plans, product)
    else
      vm.pricing_plans = plans_for_currency
      vm.orderPossible = true
```
This will ensure that any 'payg' or 'free' plans are displayed regardless of what currency is set.
I think this is going to resolve a few pricing issues 🥇

ouranos:
@NickenChuggets please have a look at new commits

ouranos:
Squashed commits before merging

PR #500, opened 2018-05-25
[MNOE-1137] Company Cart System: Issue with Modifying subscription within the cart

MAhsenArif:


src/locales/en.json
```diff
@@ -717,6 +717,7 @@
   "mno_enterprise.templates.dashboard.provisioning.confirm.name": "Name",
   "mno_enterprise.templates.dashboard.provisioning.confirm.description": "Description",
   "mno_enterprise.templates.dashboard.provisioning.confirm.edit_order": "Edit Order",
+  "mno_enterprise.templates.dashboard.provisioning.confirm.update_cart": "Update Cart",
```
ouranos: @MAhsenArif Was this added to mno-enterprise?

MAhsenArif: I think it got missed. Adding it quickly

PR #501, opened 2018-05-25
[MNOE-1135] Made address fields in company settings mandatory

RidaAhmad:


RidaAhmad:
Yeah, the condition on the address is missing, I'll add if it's required in customer view as well, thanks.

RidaAhmad:
Feedback incorporated, thank you.

PR #503, opened 2018-05-29
[MNOE-1037]on company view page - tab lost on refresh 

jin521:


src/app/views/company/company.controller.coffee
```diff
@@ -17,7 +17,8 @@ angular.module 'mnoEnterpriseAngular'
         if vm.isBillingShown()
           vm.activeTab = 'billing'
```
yuritomanek: Can this be moved into the `setActiveTab()` method to keep the logic all together?

src/app/views/company/company.controller.coffee
```diff
@@ -17,7 +17,8 @@ angular.module 'mnoEnterpriseAngular'
...
         else
-          vm.activeTab = 'members'
+      #On refreshing the page it should stay on the same tab
```
ouranos: Use one space between the leading # character of the comment and the text of the comment.

src/app/views/company/company.controller.coffee
ouranos: Agreed

PR #504, opened 2018-05-30
[MNOE-1152] Display Custom Data on Confirm

paulstuartparker:


src/app/views/provisioning/confirm.html
```diff
@@ -71,6 +71,16 @@ <h1 translate>mno_enterprise.templates.dashboard.provisioning.confirm.title</h1>
       </div>
     </div>
 
+    <!-- Display Read-Only Custom Form Data If It Exists -->
+    <div ng-show="vm.subscription.custom_data && vm.subscription.product.custom_schema">
```
ouranos: should this be an `ng-if`?

src/app/views/provisioning/confirm.controller.coffee
```diff
@@ -16,6 +16,16 @@ angular.module 'mnoEnterpriseAngular'
       editAction: $stateParams.editAction,
       cart: $stateParams.cart
 
+    setCustomSchema = () ->
+      vm.model = vm.subscription.custom_data || {}
```
ouranos: See @iseessel comment on admin panel, do we need `vm.model`?

PR #507, opened 2018-05-30
[MNOE-1014] Cart access requires app management and provisioning to be enabled

rheasunshine:
@ouranos @adamaziz15   
Solves unauthorized requests to subscriptions endpoint on layout page,  not specifically related to 1014, but cause of the defect


ouranos:
@adamaziz15 @MAhsenArif I believe this can be closed as it's been integrated in #510?

MAhsenArif:
@ouranos Yes this can be closed.

PR #510, opened 2018-06-04
[MNOE-1159] Prevent member users from requesting cart subscriptions

MAhsenArif:


src/app/components/dashboard-company-cart/dashboard-company-cart.coffee
```diff
@@ -2,20 +2,36 @@
 #============================================
 #
 #============================================
-DashboardCompanyCartCtrl = ($scope, MnoeProvisioning) ->
+DashboardCompanyCartCtrl = ($scope, MnoeProvisioning, MnoeOrganizations, MnoeCurrentUser) ->
   'ngInject'
 
-  cartSubscriptions = ->
+  initCartSubscritpions = ->
```
ouranos: Typo

MAhsenArif:
@ouranos We will still need to use the admin check in `DashboardCompanyCartCtrl` since we also want to refresh the cart subscriptions whenever an organization is changed. The check in `LayoutController`(in https://github.com/maestrano/mno-enterprise-angular/pull/507) can be useful to prevent the `DashboardCompanyCartCtrl` from being called for members, but it will not be able to refresh the cart subscriptions.

ouranos:
But we already have a ` $scope.$watch(MnoeOrganizations.getSelectedId)` on the `LayoutController` that check the role
That's why I was suggesting  combining them

src/app/views/layout.controller.coffee
```diff
@@ -17,10 +17,13 @@ angular.module 'mnoEnterpriseAngular'
     $scope.$watch(MnoeOrganizations.getSelectedId, (newValue) ->
       MnoeCurrentUser.get().then(
         (response) ->
+          selectedOrg = _.find(response.organizations, {id: newValue})
+          isAdmin = MnoeOrganizations.role.atLeastAdmin(selectedOrg.current_user_role)
+          $scope.isCartEnabled = if isAdmin then MnoeConfig.isProvisioningEnabled() else false
```
adamaziz15: Could be refactored to:
`$scope.isCartEnabled = isAdmin && MnoeConfig.isProvisioningEnabled()`

adamaziz15: Good combining the logic from https://github.com/maestrano/mno-enterprise-angular/pull/507
and good catch on the appManagement part of it. Don't want that flag to control whether or not the cart is enabled.

MAhsenArif:
@ouranos @adamaziz15 I've tested out the working with the changes and also incorporated feedback.
There is one AngularJS problem in this though. When I switch from an organization in which user has `Admin` access to one where user has `Member` access, AngularJS throws a transition error. Vice versa doesn't happen though. 
screen shot 2018-06-13 at 8 36 07 pm

 I’ve tried looking into it but i I'm not sure what AngularJS convention breaking

adamaziz15:
@MAhsenArif I've fixed the transition superseded error. Just needed to reorganize the code a bit.

The error was being raised because the `$scope.isCartEnabled` flag was being toggled before the state change between impac/apps states - and that flag triggers a transition as it is attached to an ng-if in the view.

PR #511, opened 2018-06-04
[MNOE-1167] UI tooltip updated for item removal from cart and code cl…

MAhsenArif:
…eanup


src/app/views/provisioning/subscriptions.controller.coffee
```diff
@@ -11,40 +11,6 @@ angular.module 'mnoEnterpriseAngular'
       else
         $state.go('home.subscription', { id: subscription.id })
 
-    vm.cancelSubscription = (subscription, i) ->
```
ouranos: Does that we can no longer cancel subscriptions?
Or is it because the actual cancelling is now a modify?

In this case `MnoeProvisioning.cancelSubscription` should be removed too (and do a search to make sure it's not used anywhere)

@adamaziz15  @iseessel

MAhsenArif: @ouranos All subscription actions now go to [vm.editSubscription](https://github.com/maestrano/mno-enterprise-angular/blob/2.0/src/app/views/provisioning/subscriptions.controller.coffee#L101) method. So yes I belive this and `MnoeProvisioning.cancelSubscription` can be removed.

@adamaziz15 @iseessel If could also confirm this as well.

iseessel: @MAhsenArif @ouranos Yep that is correct -- cancelling a subscription is now a modify request that creates a subscription event with event_type as cancel. 

All of these deletions looks good -- and we can delete MnoeProvisioning.cancelSubscription

PR #513, opened 2018-06-05
[MNOE-1166] Max licenses UI improvements

iseessel:


src/app/views/provisioning/confirm.html
```diff
@@ -34,9 +34,9 @@ <h1 translate>mno_enterprise.templates.dashboard.provisioning.confirm.title</h1>
     <div ng-show="(vm.singleBilling && vm.billedLocally) || vm.subscription.product.product_type != 'application'" class="table-cell">
       <div class="row th-row hidden-xs">
         <div class="cell col-sm-3">{{ 'mno_enterprise.templates.dashboard.provisioning.confirm.name' | translate }}</div>
-        <div class="cell col-sm-2">{{ 'mno_enterprise.templates.dashboard.provisioning.confirm.description' | translate }}</div>
-        <div class="cell col-sm-2">{{ 'mno_enterprise.templates.dashboard.provisioning.confirm.licenses' | translate }}</div>
-        <div class="cell col-sm-3">{{ 'mno_enterprise.templates.dashboard.provisioning.confirm.price' | translate }}</div>
+        <div class="cell col-sm-2" ng-show="vm.subscription.product_pricing.license_based">{{ 'mno_enterprise.templates.dashboard.provisioning.confirm.licenses' | translate }}</div>
```
adamaziz15: Do we need `.product_pricing?...` here in the event that there is no product_pricing attached to the subscription?

ouranos: Shouldn't we have 12 columns in this row instead of 10?
Also we should have conditional class based on {{vm.subscription.product_pricing.license_based}} so the total is always 12.
Same problem for the table body.

src/locales/en.json
```diff
@@ -739,6 +739,7 @@
   "mno_enterprise.templates.dashboard.provisioning.summary.pricing_info.single_billing_disabled": "Subscription plan is managed by the application",
   "mno_enterprise.templates.dashboard.provisioning.summary.pricing_info.externally_managed": "Subscription plan is managed by the application and billed via Maestrano",
   "mno_enterprise.templates.dashboard.provisioning.summary.pricing_info.billed_locally": "Billing is operated via Maestrano",
+  "mno_enterprise.templates.dashboard.provisioning.summary.licenses": "Licenses",
```
ouranos: Is this locale present on `mno-enterprise`?

iseessel:
Changed column widths
Added N/A to licenses and tooltip 
Added locales here: https://github.com/maestrano/mno-enterprise/pull/740

Made corresponding changes here: https://github.com/maestrano/mnoe-admin-panel/pull/297

PR #517, opened 2018-06-08
[MNOE-1183] Sync status not showing on App Management page

MAhsenArif:


src/app/views/apps-management/apps-management.controller.coffee
```diff
@@ -1,29 +1,65 @@
...
 
       vm.dataSharingStatus = (product) ->
-        if product.sync_status?.attributes?.status
-          'Connected'
+        # If sync status is set from backend then it follows following:
+        # sync_status values:
+        #   null => disconnected
+        #   all other values => connected
+        if _.find(vm.filterSyncProductIds, (uid) -> uid == product.uid)
+          product.sync_status?.attributes?.status
```
adamaziz15: Probably don't need the guarding here since if the product is in `vm.filterSyncProductIds` then it will always have the status set.

src/app/views/apps-management/apps-management.controller.coffee
```diff
@@ -1,29 +1,65 @@
...
+            'Disconnected'
 
       vm.subscriptionStatus = (product) ->
         return product.subscription.status if product.subscription
 
       vm.appActionUrl = (product) ->
         "/mnoe/launch/#{product.uid}"
 
+      vm.initAppInstanceSync = ->
+        vm.filterSyncProductIds = _.map(_.filter(vm.products, (product) -> !product.sync_status), (prod) -> prod.uid)
```
adamaziz15: I'm thinking we should rename this to something like `vm.unsyncedProductIds` so it is clear that this is an array of product ids that do not have a sync status attached (until we fetch it from the 'app_instances_sync' endpoint)

src/app/components/mnoe-api/product-instances.svc.coffee
```diff
@@ -7,7 +7,9 @@ angular.module 'mnoEnterpriseAngular'
 
     productInstancesPromise = null
     @getProductInstances = ->
-      return productInstancesPromise if productInstancesPromise?
+      # Note: Temporarily commented out product instance cache till actual
+      #       connec implementation from APPINT-1238 for sync status is done.
+      # return productInstancesPromise if productInstancesPromise?
```
adamaziz15: I'm actually not completely sure about this part, even when APPINT-1238 is complete. Having this promise will mean that any time the status is updated, it will be out of sync until the page is refreshed. Not sure of the performance implications, but it may just be better to have this removed. @ouranos Thoughts on this?

src/app/components/mnoe-api/app-instances.svc.coffee
```diff
@@ -92,4 +92,16 @@ angular.module 'mnoEnterpriseAngular'
       else
         "INSTALLED_LAUNCH"
 
+    @getAppInstanceSync = ->
+      defer = $q.defer()
```
ouranos: I don't think we need to use `$q.defer` here. And the nested promise is just resolving the response.

> If the returned value is a promise, then the further execution is suspended until it settles. After that, the result of that promise is given to the next .then handler.
-- _https://javascript.info/promise-chaining#returning-promises_

So we could just have:
```coffeescript
@getAppInstanceSync = ->
  MnoeOrganizations.get().then(
   -> _self.appInstancesSyncPromise = MnoeApiSvc.one('organizations', MnoeOrganizations.selectedId).one('/app_instances_sync').get()
  )
```
Disclaimer: double check that it works as advertised  ;)

src/app/components/mnoe-api/product-instances.svc.coffee
ouranos: Looks like this is only called on the App Management page.
So I wouldn't be shocked to have the page reload status (especially when we have a loader)
This pattern was mostly done for the `AppInstances` in the dock who took a while to load

Btw there's also `LocalStorage` caching.

src/app/views/apps-management/app-management/app-management.controller.coffee
```diff
@@ -113,6 +107,24 @@ angular.module 'mnoEnterpriseAngular'
...
           AppSettingsHelper.addOnSettingLauch(vm.product)
 
+        setSyncStatusValue = ->
+          vm.product.sync_status = _.find(vm.connec_apps, (app) -> app.uid == vm.product.uid)
+          unless vm.product.sync_status?.status
+            vm.dataSharingStatus = 'Disconnected'
+            return
+
+          vm.product.sync_status.attributes = {}
+          vm.product.sync_status.attributes.finished_at = vm.product.sync_status.last_sync_date
```
ouranos: You could combine this with the line above.

src/app/views/apps-management/apps-management.controller.coffee
ouranos: Also don't use lodash `map` and `filter` but [coffescript idioms](https://arcturo.github.io/library/coffeescript/04_idioms.html):
```coffeescript
vm.unsyncedProductIds = (prod.uid for prod in vm.products when !prod.sync_status)
```
Better perf (as it compiles to native JS  loop) and much easier to read!

src/app/components/mnoe-api/product-instances.svc.coffee
adamaziz15: Ah, removing the LocalStorage caching as well.

src/app/components/mnoe-api/product-instances.svc.coffee
```diff
@@ -7,24 +7,9 @@ angular.module 'mnoEnterpriseAngular'
 
     productInstancesPromise = null
```
iseessel: I think we can remove this now that we aren't cacheing.

adamaziz15: Ah yes, good catch 👍

adamaziz15: Adding the cache back in as the workaround implemented no longer relies on the product instances themselves

src/app/components/mnoe-api/app-instances.svc.coffee
```diff
@@ -92,4 +92,9 @@ angular.module 'mnoEnterpriseAngular'
       else
         "INSTALLED_LAUNCH"
 
+    @getAppInstanceSync = ->
+      MnoeOrganizations.get().then(
+        ->  _self.appInstancesSyncPromise = MnoeApiSvc.one('organizations', MnoeOrganizations.selectedId).one('/app_instances_sync').get()
```
iseessel: I don't think we are using `appInstancesSyncPromise ` here?

iseessel: `_self.appInstancesSyncPromise ?= .... `

PR #518, opened 2018-06-12
[MNOE-1149] add tooltips and improve bootstrap for subscription tables

jin521:
screen shot 2018-06-22 at 5 02 16 pm
screen shot 2018-06-22 at 5 01 57 pm

### DIFF
--- a/src/app/views/provisioning/order.html
+++ b/src/app/views/provisioning/order.html
@@ -12,12 +12,12 @@
       </div>
 
       <div ng-hide="vm.isLoading" class="table-cell" ng-show="!isLoading">
-        <div class="row th-row hidden-xs">
-          <div class="col-xs-10 col-xs-offset-2">
-            <div class="cell col-sm-2">{{ 'mno_enterprise.templates.dashboard.provisioning.order.name' | translate }}</div>
-            <div class="cell col-sm-4">{{ 'mno_enterprise.templates.dashboard.provisioning.order.description' | translate }}</div>
-            <div class="cell col-sm-2">{{ 'mno_enterprise.templates.dashboard.provisioning.order.price' | translate }}</div>
-            <div class="cell col-sm-3">
+        <div class="row th-row">
+          <div class="col-xs-11 col-xs-offset-1">
+            <div class="cell col-sm-3 hidden-xs" >{{ 'mno_enterprise.templates.dashboard.provisioning.order.name' | translate }}</div>
+            <div class="cell col-sm-4 hidden-xs">{{ 'mno_enterprise.templates.dashboard.provisioning.order.description' | translate }}</div>
+            <div class="cell col-sm-3 hidden-xs">{{ 'mno_enterprise.templates.dashboard.provisioning.order.price' | translate }}</div>
+            <div class="cell col-sm-2 col-xs-6">
               <select class="form-control" ng-if="vm.isCurrencySelectionEnabled" ng-model="vm.selectedCurrency" ng-options="c for c in vm.currencies" ng-change="vm.filterPricingPlans()">
               </select>
             </div>
@@ -31,12 +31,12 @@
         </div>
 
         <div class="row td-row" ng-repeat="pricingPlan in vm.filteredPricingPlans" ng-click="vm.selectPlan(pricingPlan)">
-          <div class="cell radio-selector col-xs-2">
+          <div class="cell radio-selector col-xs-1">
             <input type="radio" name="optradio" ng-model="vm.subscription.product_pricing.id" ng-value="pricingPlan.id" required>
           </div>
 
-          <div class="col-xs-10">
-            <div class="cell col-sm-2" uib-tooltip="{{ pricingPlan.name }}">
+          <div class="col-xs-11">
+            <div class="cell col-sm-3" uib-tooltip="{{ pricingPlan.name }}">
               <span class="cell-label visible-xs-inline">
                 {{ 'mno_enterprise.templates.dashboard.provisioning.order.name' | translate }}:
               </span>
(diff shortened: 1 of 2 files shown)


src/app/views/provisioning/order.html
```diff
@@ -12,12 +12,12 @@ <h1 translate>{{ vm.subscriptionPlanText }}</h1>
...
 
       <div ng-hide="vm.isLoading" class="table-cell" ng-show="!isLoading">
-        <div class="row th-row hidden-xs">
-          <div class="col-xs-10 col-xs-offset-2">
-            <div class="cell col-sm-2">{{ 'mno_enterprise.templates.dashboard.provisioning.order.name' | translate }}</div>
-            <div class="cell col-sm-4">{{ 'mno_enterprise.templates.dashboard.provisioning.order.description' | translate }}</div>
-            <div class="cell col-sm-2">{{ 'mno_enterprise.templates.dashboard.provisioning.order.price' | translate }}</div>
-            <div class="cell col-sm-3">
+        <div class="row th-row">
+          <div class="col-xs-11 col-sm-offset-1">
```
ouranos: Shouldn't they use the same breakpoint? (xs/sm)
Also to nest columns, I think we should had a `row` to wrap all the columns below.
See [Nesting columns](https://getbootstrap.com/docs/3.3/css/#grid-nesting)

src/app/views/provisioning/order.html
```diff
@@ -31,26 +31,26 @@ <h1 translate>{{ vm.subscriptionPlanText }}</h1>
...
         <div class="row td-row" ng-repeat="pricingPlan in vm.filteredPricingPlans" ng-click="vm.selectPlan(pricingPlan)">
-          <div class="cell radio-selector col-xs-2">
+          <div class="cell radio-selector col-xs-1">
             <input type="radio" name="optradio" ng-model="vm.subscription.product_pricing.id" ng-value="pricingPlan.id" required>
           </div>
 
-          <div class="col-xs-10">
-            <div class="cell col-sm-2" uib-tooltip="{{ pricingPlan.name }}">
+          <div class="col-xs-11">
+            <div class="cell col-sm-3" uib-tooltip="{{ pricingPlan.name }}">
```
ouranos: Same here, I think we need a `row` around the nested columns

Also I only count 10 columns being used?

src/app/views/provisioning/order.html
```diff
@@ -12,12 +12,12 @@ <h1 translate>{{ vm.subscriptionPlanText }}</h1>
...
-        <div class="row th-row hidden-xs">
-          <div class="col-xs-10 col-xs-offset-2">
-            <div class="cell col-sm-2">{{ 'mno_enterprise.templates.dashboard.provisioning.order.name' | translate }}</div>
-            <div class="cell col-sm-4">{{ 'mno_enterprise.templates.dashboard.provisioning.order.description' | translate }}</div>
-            <div class="cell col-sm-2">{{ 'mno_enterprise.templates.dashboard.provisioning.order.price' | translate }}</div>
-            <div class="cell col-sm-3">
+        <div class="row th-row">
+          <div class="col-xs-11 col-xs-offset-1">
+            <div class="cell col-sm-2 hidden-xs" >{{ 'mno_enterprise.templates.dashboard.provisioning.order.name' | translate }}</div>
+            <div class="cell col-sm-4 col-xs-offset-1 hidden-xs">{{ 'mno_enterprise.templates.dashboard.provisioning.order.description' | translate }}</div>
```
ouranos: Why do we have `col-xs-offset-1` if it's `hidden-xs`?

src/app/views/provisioning/order.html
```diff
@@ -12,12 +12,12 @@ <h1 translate>{{ vm.subscriptionPlanText }}</h1>
...
+            <div class="cell col-sm-3 hidden-xs">{{ 'mno_enterprise.templates.dashboard.provisioning.order.price' | translate }}</div>
+            <div class="cell col-xs-6 col-sm-2">
```
ouranos: Please keep the order consistent betweeen `sm` and `xs`.
Previous columns have `sm` before `xs`

src/app/views/provisioning/order.html
```diff
@@ -12,12 +12,12 @@ <h1 translate>{{ vm.subscriptionPlanText }}</h1>
...
-        <div class="row th-row hidden-xs">
-          <div class="col-xs-10 col-xs-offset-2">
-            <div class="cell col-sm-2">{{ 'mno_enterprise.templates.dashboard.provisioning.order.name' | translate }}</div>
-            <div class="cell col-sm-4">{{ 'mno_enterprise.templates.dashboard.provisioning.order.description' | translate }}</div>
-            <div class="cell col-sm-2">{{ 'mno_enterprise.templates.dashboard.provisioning.order.price' | translate }}</div>
-            <div class="cell col-sm-3">
+        <div class="row th-row">
+          <div class="col-xs-11 col-xs-offset-1">
+            <div class="cell col-sm-2 hidden-xs" >{{ 'mno_enterprise.templates.dashboard.provisioning.order.name' | translate }}</div>
+            <div class="cell col-sm-4 col-sm-offset-1 hidden-xs">{{ 'mno_enterprise.templates.dashboard.provisioning.order.description' | translate }}</div>
```
ouranos: Why do we have an offset here? It would be better to make use of the space?
Could you add an updated screenshot?

src/app/views/provisioning/order.html
ouranos: This hasn't been addressed?

ouranos: Ok 12 columns now

PR #519, opened 2018-06-12
[MNOE-1110][OPAL-431] Order details copying over

iseessel:


iseessel:
This adds ability to create subscription events directly. 
Tooltip for subscriptions without subscription details.

PR #520, opened 2018-06-13
[MNOE-1148] Fixed issues with provisioning of products for members

RidaAhmad:
Fixed issues with provisioning of local and regular products for different types of users

### DIFF
--- a/src/app/components/mno-app-install-btn/mno-app-install-btn.coffee
+++ b/src/app/components/mno-app-install-btn/mno-app-install-btn.coffee
@@ -35,6 +35,10 @@ angular.module 'mnoEnterpriseAngular'
 
       vm.canProvisionApp = false
 
+      atLeastAdmin = (user, currentOrg) ->
+        org = _.find(user.organizations, { id: currentOrg.id })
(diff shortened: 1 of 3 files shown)


src/app/components/mno-app-install-btn/mno-app-install-btn.coffee
```diff
@@ -172,7 +176,9 @@ angular.module 'mnoEnterpriseAngular'
             authorizedOrganizations = _.filter(currentUser.organizations, (org) ->
               MnoeOrganizations.role.atLeastAdmin(org.current_user_role)
             )
-            vm.canProvisionApp = !_.isEmpty(authorizedOrganizations)
+
+            organization = MnoeOrganizations.selected.organization
+            vm.canProvisionApp = atLeastAdmin(currentUser, organization)
```
adamaziz15: Rather than writing this helper function, we could just check that the selected organization is in authorizedOrganizations:
`vm.canProvisionApp = _.find(authorizedOrganizations, (org) -> org.id == organization.id)`

RidaAhmad: Sounds better, will update, thanks.

PR #521, opened 2018-06-14
[MNOE-1186] Refresh the subscriptions screen when changing company

jin521:
also for Marketplace- Re-seller products screen


src/app/views/provisioning/subscriptions.controller.coffee
```diff
@@ -37,25 +37,33 @@ angular.module 'mnoEnterpriseAngular'
...
+            currencies = _.map(subscription?.product_pricing?.prices, 'currency')
+            _.includes(currencies, vm.orgCurrency) || (subscription?.product_pricing?.pricing_type in PRICING_TYPES['unpriced'])
+          )
+      ).finally(-> vm.isLoading = false)
+
+    #====================================
+    # Post-Initialization
+    #====================================
+    $scope.$watch MnoeOrganizations.getSelectedId, (val) ->
+      vm.initialize() if val?
```
MAhsenArif: Can we also reset the UI loader here. Otherwise screen might remain frozen till the subscriptions are re-fetched for the new organization.

MAhsenArif: Sorry nevermind, I see the loader is set in the `initialize` method

PR #522, opened 2018-06-20
[MNOE-1171] Add loader when loading custom schema

NickenChuggets:


src/app/views/provisioning/details.controller.coffee
```diff
@@ -100,12 +100,18 @@ angular.module 'mnoEnterpriseAngular'
       fetchSubscription().then(fetchProduct).then(fetchCustomSchema)
         .then(() -> setCustomSchema(vm.subscription.product))
         .catch((error) ->
-          toastr.error('mno_enterprise.dashboard.provisioning.subscriptions.product_error')
+          toastr.error('mno_enterprise.templates.dashboard.provisioning.subscriptions.product_error')
           $state.go('home.subscriptions', {subType: if urlParams.cart then 'cart' else 'active'})
         )
         .finally(() -> vm.isLoading = false)
     else
+      vm.isLoading = true
```
adamaziz15: I think we can pull this out of the conditionals to avoid duplication

PR #531, opened 2018-07-16
[OPAL-513] Locales displayed on "Validate"

adamaziz15:
Add guarding to summary controller to prevent error thrown when product
is set to skip price selection.

### DIFF
--- a/src/app/views/provisioning/summary.controller.coffee
+++ b/src/app/views/provisioning/summary.controller.coffee
@@ -8,20 +8,22 @@ angular.module 'mnoEnterpriseAngular'
     vm.selectedCurrency = MnoeProvisioning.getSelectedCurrency()
     vm.subType = if $stateParams.cart == 'true' then 'cart' else 'active'
     vm.subscription = MnoeProvisioning.getCachedSubscription()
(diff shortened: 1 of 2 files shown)


src/app/views/provisioning/summary.controller.coffee
```diff
@@ -8,20 +8,22 @@ angular.module 'mnoEnterpriseAngular'
...
     else
       MnoeProvisioning.initSubscription({subscriptionId: $stateParams.subscriptionId})
         .then((response) ->
           vm.subscription = response
-          vm.singleBilling = vm.subscription.product.single_billing_enabled
-          vm.billedLocally = vm.subscription.product.billed_locally
-          vm.quoteBased = vm.subscription.product_pricing.quote_based
-          )
-        .finally(() -> vm.isLoading = false)
+          initSummary()
```
iseessel: Do we need to fetch the quote if it is not cached? e.g. if we refresh the page?

adamaziz15: I don't think we want to do that as you can hard code the url for the summary page. If the pricing changed since the last time they viewed the summary page we don't want them to see a different quote.

PR #538, opened 2018-08-15
[OPALSUP-265] Fix public marketplace view

jlove-nz:
The public marketplace view on the landing page now correctly
shows local products flagged as public, removes categories
where no public products are configured and on initial view
does no longer shows all products initially.

This change also fixes the category selection in the marketplace
which was not working because not all products have categories.

### DIFF
--- a/src/app/components/mno-apps/mno-app-listing.coffee
+++ b/src/app/components/mno-apps/mno-app-listing.coffee
@@ -40,7 +40,7 @@ angular.module 'mnoEnterpriseAngular'
         vm.searchTerm = ''
         vm.currentSelectedCategory = if vm.publicPage then vm.selectedPublicCategory.label else vm.selectedCategory
         if (vm.currentSelectedCategory?.length > 0)
-          vm.filteredApps = (app for app in vm.apps when vm.currentSelectedCategory in app.categories)
+          vm.filteredApps = _.filter(vm.apps, (app) -> _.includes(app.categories, vm.currentSelectedCategory))
         else
           vm.filteredApps = vm.apps
 
@@ -84,11 +84,12 @@ angular.module 'mnoEnterpriseAngular'
             response = response.plain()
 
             vm.categories = response.categories
-            vm.publicCategories = _.map(response.categories, (c) -> {label: c, active: ''})
             vm.apps = response.apps
-            vm.filteredApps = vm.apps
             if vm.publicPage
               vm.apps = _.filter(vm.apps, (app) -> _.includes(MnoeConfig.publicApplications(), app.nid))
+              categoriesWithProduct = _.uniq(_.flatten(_.map(vm.apps, (app) -> app.categories)))
+              vm.publicCategories = _.map(categoriesWithProduct, (c) -> {label: c, active: ''})
+            vm.filteredApps = vm.apps
       ).finally(-> vm.isLoading = false)
 
       $scope.$watch MnoeOrganizations.getSelectedId, (val) ->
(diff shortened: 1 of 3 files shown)


src/app/components/mno-apps/mno-app-listing.coffee
```diff
@@ -84,11 +84,12 @@ angular.module 'mnoEnterpriseAngular'
             response = response.plain()
 
             vm.categories = response.categories
-            vm.publicCategories = _.map(response.categories, (c) -> {label: c, active: ''})
             vm.apps = response.apps
-            vm.filteredApps = vm.apps
             if vm.publicPage
               vm.apps = _.filter(vm.apps, (app) -> _.includes(MnoeConfig.publicApplications(), app.nid))
+              categoriesWithProduct = _.uniq(_.flatten(_.map(vm.apps, (app) -> app.categories)))
```
ouranos: You can use the `_.property` callback shorthand:
```coffeescript
_.map(vm.apps, 'categories')
```

ouranos: I realise all this filtering for public apps is done in the frontend.
Would it be better to move it to the backend?

We already have the categories filtering logic:
* https://github.com/maestrano/mno-enterprise/blob/4.0/api/lib/mno_enterprise/concerns/controllers/jpi/v1/marketplace_controller.rb#L36
* https://github.com/maestrano/mno-enterprise/blob/4.0/core/app/models/mno_enterprise/app.rb#L16

src/app/components/mno-local-products/mno-local-product-listing.coffee
```diff
@@ -33,7 +33,7 @@ angular.module 'mnoEnterpriseAngular'
       # Post-Initialization
       #====================================
       $scope.$watch MnoeOrganizations.getSelectedId, (val) ->
-        vm.initialize() if val?
+        vm.initialize()
```
ouranos: @adamaziz15 any idea why the `if val?` was here? Anything to do with `$watch`?
I've seen this pattern a couple times

src/app/views/public/landing/landing.controller.coffee
```diff
@@ -4,26 +4,23 @@ angular.module 'mnoEnterpriseAngular'
...
-        # Products does not have a picture
+        # Select appropriate image for carousel - logo or picture
         picture = if app.pictures then app.pictures[0] else app.logo
         {
           "background-image": "url(#{picture})"
         }
 
       MnoeMarketplace.getApps().then(
         (response) ->
-          vm.products = _.filter(response.apps, (app) -> _.includes(MnoeConfig.publicApplications(), app.nid))
```
ouranos: Is this not needed?

jlove-nz: As far as I can tell, there is no usages of the products list in the landing view. The products lists have been encapsulated in to the app and local product listing components.

The only thing that the landing page has direct responsibility to provide is the highlighted product list.

src/app/components/mno-apps/mno-app-listing.coffee
jlove-nz: It is certainly a reasonable option. It'd remove the filtering in three locations (landing page, and the two app listing components).

jlove-nz: r.e. the property callback.  I wasn't aware of the shorthand. Thanks for pointing it out.

src/app/components/mno-local-products/mno-local-product-listing.coffee
jlove-nz: @adamaziz15  I have tested this scenario locally multiple times, so I'm sure it doesn't. I'll test in UAT for confirmation too.

PR #539, opened 2018-08-15
[OPALSUP-293] Application dock not updated for Member user

MAhsenArif:


src/app/components/mnoe-api/organizations.svc.coffee
```diff
@@ -26,6 +26,17 @@ angular.module 'mnoEnterpriseAngular'
       )
 
     organizationPromise = null
+
+    @reloadCurrentOrganization = () ->
+      orgId = _self.selectedId
+      $cookies.remove("#{orgId}_dhb_ref_id")
```
adamaziz15: Will this have an effect on dashboards for the organization at all?

MAhsenArif: @adamaziz15 If the user role is being changed from `Admin` to `Manager` then the impac dashboards would reload yes.
Changing the role to `Member` would not have any effect though since that doesn't contain dashboards.

PR #540, opened 2018-08-16
[MNOE-1277] Product display allow undefined org

jlove-nz:
The detailed product display component is designed to show
product information within the landing page where a user
is not logged and there is no selected organisation.

Fix the logic to support this use case by not relying on
a selected organisation.

### DIFF
--- a/src/app/components/mno-apps/mno-app.coffee
+++ b/src/app/components/mno-apps/mno-app.coffee
@@ -55,7 +55,7 @@ angular.module 'mnoEnterpriseAngular'
         vm.app ||= _.findWhere(apps, { id:  appId })
 
         # Init currency
-        vm.currency = MnoeOrganizations.selected.organization.billing_currency || MnoeConfig.marketplaceCurrency()
+        vm.currency = MnoeOrganizations.selected?.organization?.billing_currency || MnoeConfig.marketplaceCurrency()
 
         # Init Pricing Plans
         getPricingPlans()


src/app/components/mno-apps/mno-app.coffee
```diff
@@ -55,7 +55,7 @@ angular.module 'mnoEnterpriseAngular'
         vm.app ||= _.findWhere(apps, { id:  appId })
 
         # Init currency
-        vm.currency = MnoeOrganizations.selected.organization.billing_currency || MnoeConfig.marketplaceCurrency()
+        vm.currency = (if MnoeOrganizations.selected? then MnoeOrganizations.selected.organization.billing_currency else MnoeConfig.marketplaceCurrency())
```
adamaziz15: For this one I think it might be better to do something like this:
```
vm.currency = MnoeOrganizations.selected?.organization?.billing_currency || MnoeConfig.marketplaceCurrency()
```

This way, the currency will still fallback to `MnoeConfig.marketplaceCurrency` in the event that there is a selected organization but no billing_currency defined.

adamaziz15: @ouranos Actually, would it be necessary to guard against this since there should always be a billing currency set?

ouranos: Happy with what you suggest

jlove-nz: @adamaziz15 I've made the change as suggested. Cheers.

PR #541, opened 2018-08-28
[MNOE-1280] [WIP] Cleared instances cache to refresh apps & products

RidaAhmad:
Emptied out apps & products instances' cache upon subscription's validation so the instances are fetched again and hence the updated apps are displayed on dock & app management page.

### DIFF
--- a/src/app/views/provisioning/confirm.controller.coffee
+++ b/src/app/views/provisioning/confirm.controller.coffee
@@ -1,5 +1,5 @@
 angular.module 'mnoEnterpriseAngular'
-  .controller('ProvisioningConfirmCtrl', ($scope, $state, $stateParams, $log, MnoeOrganizations, MnoeProvisioning, MnoeAppInstances, MnoeConfig, ProvisioningHelper, schemaForm, toastr) ->
+  .controller('ProvisioningConfirmCtrl', ($scope, $state, $stateParams, $log, MnoeOrganizations, MnoeProvisioning, MnoeAppInstances, MnoeConfig, ProvisioningHelper, schemaForm, toastr, MnoeProductInstances) ->
 
     vm = this
 
@@ -28,6 +28,11 @@ angular.module 'mnoEnterpriseAngular'
         .then((schema) ->
           vm.schema = schema
         )
+
+    clearInstancesCache = ->
+      MnoeAppInstances.emptyAppInstances()
+      MnoeProductInstances.emptyProductInstances()
+      MnoeProductInstances.clearCache()
 
     vm.editOrder = (reload = true) ->
       switch $stateParams.editAction.toLowerCase()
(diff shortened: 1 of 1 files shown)


src/app/views/provisioning/confirm.controller.coffee
```diff
@@ -77,6 +82,7 @@ angular.module 'mnoEnterpriseAngular'
           MnoeProvisioning.refreshCartSubscriptions()
           $state.go("home.subscriptions", {subType: 'cart'})
         else
+          clearInstancesCache()
           # Reload dock apps
```
RidaAhmad: @MAhsenArif Please if you can confirm if we're good to remove this reload dock apps block of code as it no longer seems to be required now, will appreciate your input on this so you can confirm if it ain’t needed for cart or any other related functionalities.
Thank you,

MAhsenArif: @RidaAhmad We still need this for the impac dashboard(if it is enabled) as this code is responsible for reloading the dock apps.
It isnt, however related to the cart subscriptions reloading as that is a separate condition branch(the `if` part of this code).

RidaAhmad: Makes sense, thank you!

MAhsenArif: @RidaAhmad On a deeper look, it seems that since we have a [watcher](https://github.com/maestrano/mno-enterprise-angular/blob/2.0/src/app/components/dashboard-apps-dock/dashboard-apps-dock.coffee#L111) added in the app docks component, it does do the appInstance cache refresh on it's own. This watcher will always get triggered because the app dock component is completely removed from DOM and re-added whenever we switch to and fro from [Impac Dasboard](https://github.com/maestrano/mno-enterprise-angular/blob/2.0/src/app/views/impac/impac.html#L3).

The same behaviour is happening for the [apps dashboard](https://github.com/maestrano/mno-enterprise-angular/blob/2.0/src/app/views/apps/dashboard-apps-list.controller.coffee#L104)(tile view)

So it should be safe to remove the dock reload from here. and just keep the `clearInstancesCache()` method.

FYI @adamaziz15 @ouranos

src/app/views/provisioning/confirm.controller.coffee
```diff
@@ -1,5 +1,5 @@
 angular.module 'mnoEnterpriseAngular'
-  .controller('ProvisioningConfirmCtrl', ($scope, $state, $stateParams, $log, MnoeOrganizations, MnoeProvisioning, MnoeAppInstances, MnoeConfig, ProvisioningHelper, schemaForm, toastr) ->
+  .controller('ProvisioningConfirmCtrl', ($scope, $state, $stateParams, $log, MnoeOrganizations, MnoeProvisioning, MnoeAppInstances, MnoeConfig, ProvisioningHelper, schemaForm, toastr, MnoeProductInstances) ->
```
MAhsenArif: Line size increased. Might need to adjust indentation of the file.

RidaAhmad: Feedback incorporated, thanks.

PR #542, opened 2018-08-29
[OPALSUP-303] App Install button is conditional based on provisioning and multi-instances

MAhsenArif:
- When External Provisioning is enabled, App Install button should show conditional text based on multi-instances flag for the app

### DIFF
--- a/src/app/components/mno-app-install-btn/mno-app-install-btn.coffee
+++ b/src/app/components/mno-app-install-btn/mno-app-install-btn.coffee
@@ -33,6 +33,39 @@ angular.module 'mnoEnterpriseAngular'
           else
             "INSTALLABLE"
 
+      vm.appInstallationBtnClick = ->
+        switch vm.appInstallationStatus()
+          when 'INSTALLABLE'
+            if vm.isExternallyProvisioned
+              vm.provisionOrder()
+            else
+              vm.provisionApp()
+          when 'INSTALLED_LAUNCH' then vm.launchAppInstance()
+          when 'INSTALLED_CONNECT' then vm.connectAppInstance()
+
+      vm.provisionConditions = ->
+        if vm.appInstance
+          vm.vm.appInstallationBtnClick()
+        else
+          vm.provisionOrder()
+
+      vm.appInstallationBtnText = ->
+        switch vm.appInstallationStatus()
+          when 'INSTALLABLE'
+            if vm.isExternallyProvisioned
+              'mno_enterprise.templates.dashboard.marketplace.show.provision'
+            else
+              'mno_enterprise.templates.components.app_install_btn.start_app'
+          when 'INSTALLED_LAUNCH' then 'mno_enterprise.templates.components.app_install_btn.launch_app'
+          when 'INSTALLED_CONNECT' then 'mno_enterprise.templates.components.app_install_btn.connect_app'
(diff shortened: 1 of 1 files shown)


src/app/components/mno-app-install-btn/mno-app-install-btn.coffee
```diff
@@ -33,6 +33,39 @@ angular.module 'mnoEnterpriseAngular'
...
+            if vm.isExternallyProvisioned
+              vm.provisionOrder()
+            else
+              vm.provisionApp()
+          when 'INSTALLED_LAUNCH' then vm.launchAppInstance()
+          when 'INSTALLED_CONNECT' then vm.connectAppInstance()
+
+      vm.provisionConditions = ->
+        if vm.appInstance
+          vm.vm.appInstallationBtnClick()
```
adamaziz15: double vm?

alexmaestrano: As @adamaziz15 pointed out, should there be two `vm`?

MAhsenArif: No it should not. Fixed.

src/app/views/apps-management/app-management/app-management.controller.coffee
```diff
@@ -115,6 +115,8 @@ angular.module 'mnoEnterpriseAngular'
 
         # ********************** Initialize *********************************
         vm.init = ->
+          return $state.go('home.impac') unless MnoeConfig.isProvisioningEnabled()
```
ouranos: Should this also chek if AppManagement is enabled?

src/app/views/apps-management/apps-management.controller.coffee
```diff
@@ -27,6 +27,8 @@ angular.module 'mnoEnterpriseAngular'
         ).finally(-> vm.syncStatusesSet = true)
 
       vm.init = ->
+        return $state.go('home.impac') unless MnoeConfig.isProvisioningEnabled()
```
ouranos: Same here

src/app/views/apps-management/app-management/app-management.controller.coffee
ouranos: And rather than doing it this way, it should be done in the router (index.route.coffee) so that we use the `$urlRouterProvider.otherwise` block rather than manually changing state

ouranos:
@adamaziz15 @MAhsenArif I've fixed the routes, could you please review?
Also added a title to the single app management page.

src/app/views/apps-management/apps-management.controller.coffee
```diff
@@ -1,6 +1,6 @@
 angular.module 'mnoEnterpriseAngular'
   .controller('AppsManagementCtrl',
-    ($q, $scope, MnoeConfig, MnoeProductInstances, MnoeProvisioning, MnoeOrganizations, AppManagementHelper, MnoeAppInstances) ->
+    ($q, $scope, $state, MnoeConfig, MnoeProductInstances, MnoeProvisioning, MnoeOrganizations, AppManagementHelper, MnoeAppInstances) ->
```
adamaziz15: We don't need `$state` anymore right?

PR #547, opened 2018-09-13
[MNO-833] Security Remediation

ouranos:
* Remove deprecated gulp-utils 
* Update a bunch of npm dependencies
* Replace sprity with spritesmith and regenerate sprites
* Fix karma specs


conf/gulp.conf.js
```diff
@@ -50,7 +51,7 @@ for (const pathName in exports.paths) {
 exports.exitOnError = false;
 exports.errorHandler = function (title) {
   return function (err) {
-    gutil.log(gutil.colors.red(`[${title}]`), err.toString());
+    log(colors.red(`[${title}]`), err.toString());
```
alexmaestrano: Just making sure that brackets have to be in a string here?

src/images/sprites/mnoe-sprites.less
```diff
@@ -1,45 +1,66 @@
...
+To change `display` (e.g. `display: inline-block;`), we suggest using a common CSS class:
+
+// CSS
+.icon {
+  display: inline-block;
 }
 
+// HTML
+<i class="icon icon-home"></i>
+*/
```
alexmaestrano: 👍

src/images/sprites/mnoe-sprites.less
```diff
@@ -1,45 +1,66 @@
...
 .mnoe-icon-cc-icons-mno-cc-mastercard {
-  background-position: -0px -60px;
+  background-image: url(../images/sprites/mnoe-sprites.png);
+  background-position: -140px -78px;
   width: 46px;
   height: 30px;
 }
 .mnoe-icon-cc-icons-mno-cc-visa {
-  background-position: -0px -90px;
+  background-image: url(../images/sprites/mnoe-sprites.png);
```
alexmaestrano: Asking for an Angular learning opportunity. Can you also access assets through `@/images/sprites/mnoe-sprites.png`?

adamaziz15: I thought that was only in vue?

conf/gulp.conf.js
ouranos: the square brackets? or the back-ticks?

The back-ticks are for interpolation (see [template literals|https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals])
The square brackets were just here for still I think :)

src/images/sprites/mnoe-sprites.less
ouranos: It's TS aliases defined with the `paths` property in the `compilerOptions` of the `tsconfig.json` file.
eg:
```json
"compilerOptions": {
  "baseUrl": "./",
      "paths": {
        "@models/*": [
          "src/app/_models/*"
        ],
     ...
  }
}
```

PR #548, opened 2018-11-04
[MNOE-1297] Subs Management

adamaziz15:
- Update apps management controllers to find subscriptions using
product_instance_id


adamaziz15:
@ouranos @alexmaestrano Yep, issue was related to products that are multi-instantiable. Since we weren't matching subscriptions up with their product instance directly there were some cases where cancelled subscriptions were shown. There were also times where 2 product instances were pointing to the same subscription in apps management (we were using nid to match them up before).

adamaziz15:
Also, cherry picked that gulp commit and opened a separate PR for that (https://github.com/maestrano/mno-enterprise-angular/pull/550) so it doesn't need to wait for anything else to be merged.

